PR #4, opened 2017-04-04
Added function to determine if context manager (ctx) and run ctx as s…

AmirTadmor:
add function to determine if context manager
add function to run context manager as simple function


easypy/misc.py
```diff
@@ -35,3 +35,22 @@ def get_all_subclasses(cls):
                                    for subclass in cls.__subclasses__()
                                    for rec_subclass in get_all_subclasses(subclass)]
     return [subclass for subclass in _all if not hasattr(subclass, "_%s__is_mixin" % subclass.__name__)]
+
+
+def is_context_manager(func):
+    from contextlib import contextmanager
+
+    _contextmanager_code_object = contextmanager(None).__code__
```
koreno: This can move to the module level, no need to recreate it each time. Could also be a private param:
```
def is_context_manager(func, _cm_code_object=contextmanager(None).__code__):
...
```

easypy/misc.py
```diff
@@ -35,3 +35,22 @@ def get_all_subclasses(cls):
...
+
+    return func.__code__ == _contextmanager_code_object
+
+
+def run_as_func(ctx, *args, **kwargs):
+    if not callable(ctx):
```
koreno: naaa... it should `assert callable(ctx)`. We don't want to silence mistakes.

idanarye:
This is waaay too much work for what you want to do. I pointed you to my [black magic](https://github.com/weka-io/black-magic/blob/master/python/context_manager_aware_decorator.py) because I thought you need something similar(function that is or isn't a context manager depending on another function), but if your function is always a regular function it can be simpler:

```python
def run_as_func(ctx, *args, **kwargs):
    context_or_result = ctx(*args, **kwargs)
    if hasattr(context_or_result, '__enter__'):
        with context_or_result as result:
            pass
        return result
    else:
        return context_or_result
```

Or even:

```python
def ctx_as_value(context_or_value):
    if hasattr(context_or_value, '__enter__'):
        with context_or_result as value:
            pass
        return value
    else:
        return context_or_value
```

And use it directly on the function's returned value:

```python
foo = ctx_as_value(function_that_may_retunr_value_or_ctx())
```

easypy/misc.py
tigrawap: Name is misleading. is_context_manager this way checks only, that it is contextmanager crated with contextlib wrapper. This does not have to be true

koreno:
@idanarye  your second option is dangerous, for example if we return an object that happens to also behave as a ctx manager.

idanarye:
image
If it looks like a context manager, swims like a context manager, and quacks like a context manager - then it probably is a context manager.

easypy/misc.py
AmirTadmor: you might be right, it probably should be named `is_function_context_manager`

AmirTadmor: it can also be changed to accommodate regular context managers:
```
def is_context_manager(func):
    return (func.__code__ == _contextmanager_code_object or 
           (hasattr(func, '__enter__') and hasattr(func, '__exit__'))
```

koreno:
As discussed, checking specifically that a function was decorated by contextmanager isn't generic enough.

Since ultimately the point is to normalize a set of functions/ctxmgrs into a compatible invocation syntax, we suggest going with a `as_contextmanager` wrapper that would be applied to function only, and allow them to be used using a `with` statement.
```
actions = [cm1, cm2, as_contextmanager(func1),  as_contextmanager(func2)]
for action in actions:
    with action():
        pass
```

To be safe, `as_contextmanager` should make sure that the function does not return a cm itself, which could indicate that the user made a mistake and actually supplied a `@contextmanager` and not a function.

PR #6, opened 2017-04-06
threadtree: refactor logging's Context into ThreadContext

koreno:
feel free to suggest alternate names there

### DIFF
--- a/easypy/concurrency.py
+++ b/easypy/concurrency.py
@@ -174,7 +174,11 @@ class MultiException(PException):
                     buff.write(exc.render(**kw))
                 else:
                     if hasattr(exc, "context"):
-                        context = "(%s)" % ", ".join("%s=%s" % p for p in sorted(exc.context.items()))
+                        context = exc.context.copy()
+                        context.pop("indentation", None)
+                        if 'context' in context:  # 'context' should be renamed 'breadcrumbs'
+                            context['context'] = ";".join(context['context'])
+                        context = "(%s)" % ", ".join("%s=%s" % p for p in sorted(context.items()))
                     else:
                         context = ""
                     buff.write("{}: {}", exc, context)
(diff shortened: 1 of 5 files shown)


easypy/threadtree.py
```diff
@@ -258,3 +261,116 @@ def dump_threads():
...
+        with TC(i=1):
+            assert TC.i == 1
+            assert TC.j == 0
+
+            with TC(i=1, j=1):
+                assert TC.i == 2
+                assert TC.j == 1
+
+
+    'concatenated': attributes named here get concatenated into a list, instead of overwritten:
```
tigrawap: concatenated usually used for strings...  I'd say "merged", "joined", "appended"

easypy/threadtree.py
```diff
@@ -258,3 +261,116 @@ def dump_threads():
...
+        # (thread-b)
+        assert TC.my_data == 'b'
+        with TC(my_data='c'):
+            spawn_thread_c()
+        assert TC.my_data == 'b'
+
+        # (thread-c)
+        assert TC.my_data == 'c'
+
+    'accumulated': attributes named here get incremented each time, instead of overwritten:
```
tigrawap: 1) accumulated - incremental?
2) MB add `depth` as common incremental attribute? Sounds like often need for cases when such context would be used

koreno: 1. I could change them to `append` and `increment`
2. hmmm, we haven't needed that ourselves, so it doesn't seem like a common need. different apps might want to increase depth differently too

easypy/concurrency.py
```diff
@@ -174,7 +174,11 @@ def _get_buffer(self, **kw):
                     buff.write(exc.render(**kw))
                 else:
                     if hasattr(exc, "context"):
-                        context = "(%s)" % ", ".join("%s=%s" % p for p in sorted(exc.context.items()))
+                        context = exc.context.copy()
+                        context.pop("indentation", None)
+                        if 'context' in context:  # 'context' should be renamed 'breadcrumbs'
+                            context['context'] = "%s" % ";".join(context['context'])
```
idanarye: The `"%s" % ` is redundant

PR #9, opened 2017-04-17
Step by step

idanarye:
@koreno - I want to use this for the installation tests. Instead of using a bare generator and hoping we counted the `yield`s and and `next` correctly and in sync, a `step_by_step` will allow us to enforce that the installation step the test thinks we are at is the same step the infra says we are at.

### DIFF
--- a/easypy/lockstep.py
+++ b/easypy/lockstep.py
@@ -1,0 +1,106 @@
+from contextlib import contextmanager
+from functools import wraps
+
+from .exceptions import TException
+
+
+class LockstepSyncMismatch(TException):
+    template = "Expected {lockstep_name} to be {expected_step}, but it is {actual_step}"
+
+
+class _LockstepInvocation(object):
+    def __init__(self, name, generator):
+        self._name = name
+        self._generator = generator
+        self._current_step = 'just-started'
+
+    def step_next(self, step):
+        """Progress to the next step, confirming it's the specified one"""
+
+        try:
+            self._current_step, value = self._next_step_and_value()
+        except StopIteration:
+            raise LockstepSyncMismatch(lockstep_name=self._name,
+                                       expected_step=step,
+                                       actual_step='finished')
+        if self._current_step != step:
+            raise LockstepSyncMismatch(lockstep_name=self._name,
+                                       expected_step=step,
+                                       actual_step=self._current_step)
+        return value
+
+    def step_until(self, step):
+        """Progress until we get to the specified step, confirming that it exist"""
+
+        while self._current_step != step:
+            try:
+                self._current_step, value = self._next_step_and_value()
+            except StopIteration:
+                raise LockstepSyncMismatch(lockstep_name=self._name,
+                                           expected_step=step,
+                                           actual_step='finished')
+        return value
+
+    def step_all(self):
+        """Progress through all the remaining steps"""
+
+        for self._current_step in self._generator:
+            pass
+
+    def _next_step_and_value(self):
+        yield_result = next(self._generator)
+        if isinstance(yield_result, tuple) and len(yield_result) == 2:
+            return yield_result
+        else:
+            return yield_result, None
+
+    def __str__(self):
+        return '%s<%s>' % (self._name, self._current_step)
+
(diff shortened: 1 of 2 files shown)


easypy/step_by_step.py
```diff
@@ -0,0 +1,59 @@
...
+        if self._current_step != step:
+            raise StepByStepSyncMismatch(step_by_step_name=self._name,
+                                         expected_step=step,
+                                         actual_step=self._current_step)
+
+    def __str__(self):
+        return '%s<%s>' % (self._name, self._current_step)
+
+
+def step_by_step(generator_func):
```
koreno: `lockstep`?

easypy/step_by_step.py
```diff
@@ -0,0 +1,59 @@
...
+            step_not_taken = next(invocation._generator)
+        except StopIteration:
+            # all is well - all steps were exhausted
+            invocation._current_step = 'finished'
+        else:
+            raise StepByStepSyncMismatch(step_by_step_name=invocation._name,
+                                         expected_step='finished',
+                                         actual_step=step_not_taken)
+
+    return wrapper
```
koreno: `wrapper.execute` -> executes all steps

easypy/lockstep.py
```diff
@@ -0,0 +1,86 @@
...
+        else:
+            raise LockstepSyncMismatch(lockstep_name=invocation._name,
+                                       expected_step='finished',
+                                       actual_step=step_not_taken)
+
+    @wraps(generator_func)
+    def wrapper(*args, **kwargs):
+        context = cm(*args, **kwargs)
+
+        def execute():
```
idanarye: @koreno: This is a bit hacky, but I had to do it to implement the `execute` method as an attribute of the context object. I'm not sure we'll use it though - I think it would be better to create wrapper methods:

```python
@lockstep
def clusterize_lockstep(self, ...):
    # ...

def clusterize(self, ...):
    with self.clusterize_lockstep(...) as process:
        process.exhaust()
```

I put this feature is in a separate commit so that we can easily remove it if we want.

tests/test_lockstep.py
```diff
@@ -0,0 +1,131 @@
...
+        nonlocal finished
+
+        yield 'STEP_1'
+        yield 'STEP_2'
+        yield 'STEP_3'
+
+        finished = True
+
+    assert not finished
+    process().execute()
```
koreno: what I was thinking is `process.execute()`

easypy/lockstep.py
koreno: what I was thinking is `process.execute()`, not `process().execute()`, which I think simplifies.

not sure what you meant with "create wrapper methods"

easypy/lockstep.py
```diff
@@ -0,0 +1,104 @@
...
+            invocation._current_step = 'finished'
+        else:
+            raise LockstepSyncMismatch(lockstep_name=invocation._name,
+                                       expected_step='finished',
+                                       actual_step=step_not_taken)
+
+    @wraps(generator_func)
+    def wrapper(*args, **kwargs):
+        result = cm(*args, **kwargs)
+        result.__class__ = _LockstepGeneratorContextManager
```
idanarye: @koreno - sorry for this ugly hack, but...

Somewhere between Python 3.4.2 and 3.4.5 they changed `_GeneratorContextManager.__init__` from
```python
class _GeneratorContextManager(ContextDecorator):
    """Helper for @contextmanager decorator."""

    def __init__(self, func, *args, **kwds):
```

to
```python
class _GeneratorContextManager(ContextDecorator):
    """Helper for @contextmanager decorator."""

    def __init__(self, func, args, kwds):
```

This means we can't use the same code to initiate both versions. Well, it was a private interface to begin with...

So instead, I'll use `contextmanager` to create the object, and switch it's class to something that inherits what `contextmanager` usually returns.

koreno: must we switch the class? isn't it more polite to wrap it?

idanarye: Can do

PR #11, opened 2017-04-24
Add `deprecated_arguments` decorator.

idanarye:
We can use `@deprecated` to safely rename functions and fix calls to
them based on the deprecation warnings. This decorator does the same
thing for function arguments.

### DIFF
--- a/easypy/decorations.py
+++ b/easypy/decorations.py
@@ -22,6 +22,42 @@ def deprecated(func=None, message=None):
     return inner
 
 
+def deprecated_arguments(**argmap):
+    def wrapper(func):
+        """
+        Renames arguments while emitting deprecation warning
+
+        @deprecated_arguments(old_name='new_name')
+        def func(new_name):
+            # ...
+
+        func(old_name='value meant for new name')
+        """
+
+        @wraps(func)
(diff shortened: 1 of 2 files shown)


idanarye:
@leekeren is going to love this

koreno:
- You say "rename functions"?
- Can you clarify the use case? in other words, when should I use this and not just go through the code and refactor it completely?

Bonus question
- How can we connect the deprecation warning into some centralized logging (slack) so we can effectively clean up the mess behind us eventually?

idanarye:
* _Rename function_: say we have
    ```python
    def foo():
        pass
    ```
    If we want to rename `foo` to `bar`:
    ```python
    def bar():
        pass

    @deprecated
    def foo():
        return bar()
    ```
* Example use-case: in our new cloud events the event type is renamed from `code` to `type`. I used that renaming technique on the `Event` class, but I want the events tracker expector registration method to reflect that too - use the argument `expected_types` instead of `expected_codes`. I don't want to break code that I happen to miss though, so I'd rather use this decorator.
* Connecting the deprecation warnings to Slack is orthogonal - it's something we want anyways for ol' `@deprecated`.

PR #16, opened 2017-04-28
support for generators and ctx-manager with logging contexts

koreno:


easypy/contexts.py
```diff
@@ -16,3 +20,85 @@ def enter_context(self, *key):
...
+                    with func(*args, **kwds) as ret:
+                        yield ret
+        else:
+            def inner(*args, **kwds):
+                with self._recreate_cm():
+                    return func(*args, **kwds)
+        return wraps(func)(inner)
+
+
+if signature(_GeneratorContextManager).parameters['args'].kind is Parameter.VAR_POSITIONAL:
```
idanarye: Would have been less atrocious to monkey-patch `_GeneratorContextManager`...

At least add some comments that you do this to deal with internal changes in Python's standard library...

koreno:
Added the comments. Not sure how you mean to monkey patch and how that would be less atrocious... you're welcome to suggest an edit.

idanarye:
This is what I mean: : #18 

It's less atrocious because the internal API it depends on is less change-prune(changing how it's called makes sense. Changing the name... less), and that didn't(as far as we know) change so we don't need to account for two different styles.

PR #17, opened 2017-04-30
Timed cache

noopti:
* add support for real threads to run concurrently with gevent greenlets and monkey patching.
* timecache: allow use of arbitrary function to fetch a "timestamp"


easypy/caching.py
```diff
@@ -179,12 +180,14 @@ def inner(*args, **kwargs):
...
-                    if result is not NOT_FOUND and time.time() - ts > expiration:
+                    if result is not NOT_FOUND and get_ts_func() - ts > expiration:
                         result = NOT_FOUND
                         del cache[key]
 
                 if result is NOT_FOUND:
-                    ts = time.time()
+                    if log_recalculation:
+                        _logger.debug('time cache expired, calculating new value for %s', name)
+                    ts = get_ts_func()
```
koreno: I suspect you want to call `get_ts_func` only once per invocation. `now = get_ts_func()`

easypy/caching.py
```diff
@@ -159,7 +159,7 @@ def clear():
     return deco
 
 
-def timecache(expiration=0, typed=False):
+def timecache(expiration=0, typed=False, get_ts_func=time.time, log_recalculation=False):
```
koreno: `get_ts_func` -> `time_func`

easypy/concurrency.py
```diff
@@ -377,22 +390,31 @@ def kill(wait=False):
...
-            else:
-                if future:
-                    future.result()
-        except FutureTimeoutError as exc:
-            if not timeout:
-                # ??
-                raise
-            futures.kill()
-            _logger.warning("Concurrent future timed out (%s)", exc)
+    batch_size = kw.pop("batch_size", CONCURRENT_FIND_DEFAULT_BATCH_SIZE)
```
koreno: why not keep the `workers` parameter name?

easypy/concurrency.py
```diff
@@ -377,22 +390,31 @@ def kill(wait=False):
 
 def concurrent_find(func, params, **kw):
     timeout = kw.pop("concurrent_timeout", None)
-    with async(func, list(params), **kw) as futures:
```
koreno: is this refactoring actually needed? `async` takes a `workers` parameter, and if you break out in the middle it should terminate the executor and the remaining tasks...

easypy/concurrency.py
```diff
@@ -784,6 +806,8 @@ def __init__(self, func, *args, **kwargs):
         self.sleep = kwargs.pop('sleep', 1)
         self.loop = kwargs.pop('loop', False)
         self.timer = None
+        # in case of using apply_gevent_patch function - use this option in order to defer some jobs to real threads
+        self.real_thread_no_greenlet = kwargs.pop('real_thread_no_greenlet', False)
```
koreno: hmm... sounds goofy...
maybe "os_thread", "real_thread"?

easypy/concurrency.py
```diff
@@ -831,22 +865,29 @@ def paused(self):
...
             self._logged_func()
             yield self
             return
 
-        self.thread.start()
+        if not self.real_thread_no_greenlet:
+            _logger.debug('sending job to gevent thread')
+            self.thread.start()
+        else:
+            _logger.debug('sending job to non patched thread')
```
koreno: `'sending job to an os thread'`

easypy/concurrency.py
```diff
@@ -931,14 +972,15 @@ class RWLock(object):
 
     """
 
-    def __init__(self):
+    def __init__(self, name=None):
```
koreno: yep, good idea

easypy/gevent_non_patched_threadpool.py
```diff
@@ -0,0 +1,36 @@
+from logging import getLogger
+from queue import Queue
+import threading
+
+
+_logger = getLogger()
+non_patched_threadpool = None
+
+
+class NonPatchedThreadPool:
```
koreno: This can't be a standalone module. 
Either move this into gevent.py, or make a `gevent` sub-package with this and more inside...

easypy/properties.py
```diff
@@ -41,3 +41,7 @@ def callable(*args, **kwargs):
         return property(callable, fset, fdel, doc)
     else:
         return property(fget, fset, fdel, doc)
+
+
+def add_property(obj, name, function):
+    setattr(obj.__class__, name, property(function))
```
koreno: you can't have such a function, that takes a _specific_ object instance and sets to its not-so-specific _class_ a _specific_ function...

easypy/caching.py
idanarye: Different approach - I would have called it **after** running the function. Since it's a cache we should assume `func()` may take a while, so it makes a difference...

noopti: @idanarye - what do you mean? how is it possible to call time_func after you run the function? you have to run it before in order to determine whether you should run the function at all...

easypy/concurrency.py
noopti: I think it's a bit misleading. You would expect a worker to take another job right when it finishes one, and not wait for all other workers to finish and then take the job based on their return value.

easypy/caching.py
tigrawap: What @idanarye probably means (and I agree if this what he means) - It's more correct to move `ts = get_ts_func()` under `result = func(*args, **kwargs)`

So timestamp will represent when result was generated, not when it began to generate.  

And for a same reason it's incorrect to put `now = get_ts_func()` in the beginning of function. Validation and generation timestamps might differ

easypy/concurrency.py
tigrawap: Hm, what's a reason to change behaviour to be batched?

easypy/caching.py
idanarye: I'm referring to the case of calling `get_ts_func()` it twice:
```python
if expiration > 0:
    # First call to get_ts_func() to check if we need to run func()
    if result is not NOT_FOUND and get_ts_func() - ts > expiration:
        result = NOT_FOUND
        del cache[key]

if result is NOT_FOUND:
    if log_recalculation:
        _logger.debug('time cache expired, calculating new value for %s', name)
    result = func(*args, **kwargs)
    # Second call to get_ts_func() to determine the cache time
    ts = get_ts_func()
    cache[key] = result, ts
```
The main point of this approach is that we measure the expiration time from *after* `func()` gets called. So, for example, if `expiration=10` and `func()` takes 5 seconds to run, in the current implementation:

- t+0: Call `func()`. Set cache expiration time to `t+10`
- t+5: `func()` returns
- t+10: `func()` expires

So the result expires 5 seconds after we get it. But with the approach I described above:

- t+0: Call `func()`
- t+5: `func()` returns. Set cache expiration time to `t+15`
- t+15: `func()` expires

So the result expires 10 seconds after we get it.

noopti: ok got it.

easypy/concurrency.py
noopti: in case of running concurrent_find on a large collection, we don't want to launch an enormous amount of threads / greenlets.

easypy/threadtree.py
```diff
@@ -241,13 +234,22 @@ def watch_threads(interval):
     def dump_threads():
         nonlocal last_threads
 
+        _logger.debug('getting thread tree')
```
koreno: you can use `_logger.indented(..., level=logging.DEBUG)`

koreno:
I've removed the gevent-related changes - they'll go in a separate branch

PR #18, opened 2017-04-30
Fix #10 - loggers(any CM) can propely decorate CMs and generators

### LINKED ISSUE #10: `Logger.context` can not handle generators:
```python
In [1]: @_logger.context('my-ctx', indent='my-indent')
   ...: def my_generator():
   ...:     _logger.info('before yield')
   ...:     yield
   ...:     _logger.info('after yield')
   ...:

In [2]: for _ in my_generator():
   ...:     pass
   ...:
17:42:19|INFO1|IdanArye                 |  ├───┮ my-indent
17:42:19|INFO1|IdanArye                 |  │   ╰╼DONE in no-time (my-indent)
17:42:19|INFO |IdanArye                 |  │ before yield
17:42:19|INFO |IdanArye                 |  │ after yield
```

idanarye:
Alternative to #16 using monkey-patching

### DIFF
--- a/easypy/contexts.py
+++ b/easypy/contexts.py
@@ -1,5 +1,7 @@
 from collections import defaultdict
-from contextlib import ExitStack
+from functools import wraps
+from inspect import isgeneratorfunction
+from contextlib import ExitStack, contextmanager, _GeneratorContextManager
 
 
 class KeyedStack(ExitStack):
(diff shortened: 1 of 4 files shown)


koreno:
No no no... I don't want the responsibility, definitely not as a side-effect of importing a module from easypy.

PR #19, opened 2017-05-08
Random meaningful names

DoronCohen:
This adds `random_meaningful_name(max_length)` for a nice `adjective_adverb_verb_noun` random name (`max_length` might make the name contain less parts). Notice the dictionaries loaded here are VERY broad so the random names may contain words no one heard of. This, however, is easy to change by replacing the dictionary. The dictionaries should be sorted by word length descending.

Do we want to change `random_filename` to use it?


easypy/randutils.py
```diff
@@ -41,6 +42,48 @@ def get_chunks(self, offset, end, block_size_range, shuffle=False):
             total_size -= size
             offset += size
 
+def load_dictionary(filename):
+    full_path = os.path.join(os.path.dirname(__file__), 'dictionary', filename)
+    with open(full_path) as dict_file:
+        dictionary = dict_file.read().splitlines()
```
koreno: Use the open file as an iterator:
```dictionary = list(dict_file)```

File: easypy/randutils.py [hunk 2]
```diff
@@ -41,6 +42,48 @@ def get_chunks(self, offset, end, block_size_range, shuffle=False):
...
+
+
+NOUNS = load_dictionary('nouns.txt')
+VERBS = load_dictionary('verbs.txt')
+ADVERBS = load_dictionary('adverbs.txt')
+ADJECTIVES = load_dictionary('adjectives.txt')
+PARTS_OF_SPEECH = (NOUNS, VERBS, ADVERBS, ADJECTIVES)
+
+
+def random_meaningful_name(max_length=50):
```
koreno: you should add some unittests under ./tests/*

easypy/randutils.py
```diff
@@ -41,6 +42,48 @@ def get_chunks(self, offset, end, block_size_range, shuffle=False):
...
 
+def load_dictionary(filename):
+    full_path = os.path.join(os.path.dirname(__file__), 'dictionary', filename)
+    with open(full_path) as dict_file:
+        dictionary = dict_file.read().splitlines()
+
+    dictionary.sort(key=len)
+    return {
+        length: list(group) for length, group
+        in itertools.groupby(dictionary, key=len)
```
koreno: you can actually use `easypy.collections.grouped(..)` for this

easypy/randutils.py
```diff
@@ -41,6 +42,48 @@ def get_chunks(self, offset, end, block_size_range, shuffle=False):
...
+def random_meaningful_name(max_length=50):
+    name = ''
+    remaining = max_length
+
+    if max_length < 3:
+        return random_string(max_length, charset=string.ascii_lowercase)
+
+    for part_group in PARTS_OF_SPEECH:
+        max_part_len = remaining - 1 if name else max_length
+        part_lengths = [k for k in part_group.keys() if k <= max_part_len]
```
koreno: I'm a bit concerned with performance, when/if this is used at scale.
```
usable_word_groups = [part_group[l] for l in range(1, max_part_len)]
word_group = random.choice(usable_word_groups)
...
```

koreno: (and make `part_group` a `defaultdict(list)`)

easypy/randutils.py
```diff
@@ -41,6 +42,48 @@ def get_chunks(self, offset, end, block_size_range, shuffle=False):
...
+
+NOUNS = load_dictionary('nouns.txt')
+VERBS = load_dictionary('verbs.txt')
+ADVERBS = load_dictionary('adverbs.txt')
+ADJECTIVES = load_dictionary('adjectives.txt')
+PARTS_OF_SPEECH = (NOUNS, VERBS, ADVERBS, ADJECTIVES)
+
+
+def random_meaningful_name(max_length=50):
+    name = ''
```
koreno: would be more elegant to do:
```
parts = []
for group in PARTS_OF_SPEECH:
   ...
   parts.append(word)
return sep.join(parts)
```

easypy/randutils.py
koreno: `..., sep="-"):`

koreno:
Also, please squash all commits into one, we don't need to see the work-in-progress commits

easypy/randutils.py
DoronCohen: This will include newlines in each word.
maybe `dict_file.readlines()`?

easypy/randutils.py
DoronCohen: Oh, I didn't know `in dict.keys()` is not that efficient. I will need to filter by non empty groups though so I can detect when there are no usable word groups.
```
usable_word_groups = [
    part_group[l] for l in range(1, max_part_len)
    if len(part_group)
]
```

DoronCohen:
changed, squashed and pushed

easypy/randutils.py
```diff
@@ -42,6 +47,49 @@ def get_chunks(self, offset, end, block_size_range, shuffle=False):
...
+
+        # No available words in this group with the maximal size of
+        # max_part_len
+        if not usable_word_groups:
+            break
+
+        word_group = random.choice(usable_word_groups)
+        word = random.choice(word_group)
+        parts.append(word)
+        remaining = remaining - len(word) - (len(sep) if len(parts) else 0)
```
AmirTadmor: you can just put `len(sep) if parts else 0`

easypy/randutils.py
```diff
@@ -42,6 +47,49 @@ def get_chunks(self, offset, end, block_size_range, shuffle=False):
...
+    remaining = max_length
+
+    if max_length < 3:
+        return random_string(max_length, charset=string.ascii_lowercase)
+
+    for part_group in PARTS_OF_SPEECH:
+        max_part_len = remaining - len(sep) if parts else max_length
+        usable_word_groups = [
+            part_group[l] for l in range(1, max_part_len)
+            if len(part_group[l])
```
AmirTadmor: why `len`? `if part_group[l]` would suffice to weed out empty groups

easypy/randutils.py
```diff
@@ -42,6 +47,49 @@ def get_chunks(self, offset, end, block_size_range, shuffle=False):
...
+
+def random_meaningful_name(max_length=50, sep='-'):
+    parts = []
+    remaining = max_length
+
+    if max_length < 3:
+        return random_string(max_length, charset=string.ascii_lowercase)
+
+    for part_group in PARTS_OF_SPEECH:
+        max_part_len = remaining - len(sep) if parts else max_length
```
AmirTadmor: you will always have an extra unused char. the last part isn't followed by a seperator

easypy/randutils.py
AmirTadmor: the condition will always be True, because you check the length after appending

easypy/randutils.py
```diff
@@ -42,6 +47,49 @@ def get_chunks(self, offset, end, block_size_range, shuffle=False):
...
+        usable_word_groups = [
+            part_group[l] for l in range(1, max_part_len)
+            if len(part_group[l])
+        ]
+
+        # No available words in this group with the maximal size of
+        # max_part_len
+        if not usable_word_groups:
+            break
```
AmirTadmor: `continue`? maybe other groups do?

easypy/randutils.py
```diff
@@ -42,6 +47,49 @@ def get_chunks(self, offset, end, block_size_range, shuffle=False):
...
+
+        word_group = random.choice(usable_word_groups)
+        word = random.choice(word_group)
+        parts.append(word)
+        remaining = remaining - len(word) - (len(sep) if len(parts) else 0)
+
```
AmirTadmor: `if remaining == 0 : break`?

easypy/randutils.py
```diff
@@ -42,6 +47,49 @@ def get_chunks(self, offset, end, block_size_range, shuffle=False):
...
+    return sep.join(parts[::-1])
```
AmirTadmor: why is this inverted? a `PARTS_OF_SPEECH = (ADJECTIVES ,NOUNS, VERBS, ADVERBS)` would be more logical

easypy/randutils.py
DoronCohen: You might get wierd sentences like `beautiful_walking`. Is that ok?

easypy/randutils.py
DoronCohen: because if the max_length is reached I want to get at least a `verb_noun` or `noun` and not `adjective_adverb`. Unless defined otherwise...

easypy/randutils.py
DoronCohen: _You might get wierd sentences like beautiful_walking. Is that ok?_

From above. If we leave it that way then it is unnecessary.

easypy/randutils.py
DoronCohen: That's why I have `if parts else max_length` that way I am always calculating the prefixing separator. i.e. I do not add it for the first word.

easypy/randutils.py
AmirTadmor: but right now you will get "blue-faster-swim-man" name and that sound even worse

easypy/randutils.py
AmirTadmor: if there is no remaining then there is no point in continue iterating

easypy/randutils.py
AmirTadmor: let's do a test run; you the first three words with length (x, y, z respectively) and the separator '-' (length is 1) with max_length as 50 that means the remaining = 50 - x - 1 - y - 1 - z - 1, now you reach here and what to know how long your string can be, well parts is not empty so it's True, hence max_part_len will be remaining - 1 but there is no separator at the end of the last part, so there is no reason to decrease its size from max_part_len of the last part

easypy/randutils.py
AmirTadmor: and I didn't say to invert the tuple only to move adj to proceed the noun so this might sound sound logical

easypy/randutils.py
```diff
@@ -60,34 +60,36 @@ def load_dictionary(filename):
...
-    remaining = max_length
+    next_max_part_len = max_length
 
     if max_length < 3:
         return random_string(max_length, charset=string.ascii_lowercase)
 
-    for part_group in PARTS_OF_SPEECH:
-        max_part_len = remaining - len(sep) if parts else max_length
+    # Reverse iterate so that partial sentences will sound logical
+    for part_group in PARTS_OF_SPEECH[::-1]:
```
AmirTadmor: why not just arrange the parts of speech the way no reversing is needed?

DoronCohen: Because if the requested order is `adverb_verb_adjective_noun` partial sentences won't be logical if parts are iterated in order. We will get `adverb` or `adverb_verb_adjective`. Iterating backwards makes all possible partial sentences logical (apart from weird words chosen by random due to the extreme broadness of the dictionary).

DoronCohen:
We can squash and merge if we don't want to see all 3 commits in base.

tigrawap:
Hate to ruin your party, but this is overkill in several ways...

- Paying 100 ms to load easypy every time I use it. So dictionaries should at least be lazy. 
- Even if it's lazy - I really don't want to pay 100ms for single use of random name generation  
- Having gazillions of combinations is not something required for general use. It could be separate library. If you want it - you use it. But for general use it should be much more lightweight  
- For our specific use we also do not really need gazillions of combination. Even more, I'd prefer simplier names "dirty-jonhson" , "strange-way". Dict of 500*500 would be much more then actually needed. And should be lazily-created lists in .py, imported on function level, so no import done before actual call  
Not reading txt. This should be much faster then reading txt in python.

Default size:  
`adrenocorticotrophic-arrest-burl-cagily` - not helpful, can't say that such default is usable

Size 20
`contradistinctive` - It's definitely nice name for build server, but I can't trust myself not doing typo here

Compare this to what docker generates, which is a good example for random names:
```
laughing_lumiere
admiring_bohr
focused_leavitt
stoic_pike
stoic_mcnulty
dazzling_mahavira
priceless_cray
silly_neumann
determined_wescoff
suspicious_shirley
pedantic_stonebraker
peaceful_fermi
```

Two-words names are pretty enough. I can't imagine scenario, where I want long name with absolutely zero chances to remember it. In such case - i'd go with uuid or some random unique id.

koreno:
@DoronCohen - I agree with Anton - before this gets merged you should make the dict loading lazy, and trim the dictionaries (or find others) that generate more user-friendly names. This is ultimately what we want this for, memorable/easy-to-type strings.

DoronCohen:
I agree too. It just seemed, by talking to other people, that we need a very small chance for collision when creating a lot of files. What is a reasonable scale of combinations? are 1000^3 enough?

tigrawap:
It does not matter
Such library has no way of ensuring uniquness, so it should not care about
it all
Having big number of combinations gives false confidence, but proper way is
for application that uses names to ensure that name is valid and regenerate
if it's not

So I really not exaggerated when saying that 500*500 is enough. And it's
just 32k of occupied memory once lists are imported for a first time
(defined in function with lru_cache, file with function imported only
on-request to generate from function that generates)



2017-05-11 12:31 GMT+03:00 doronc <notifications@github.com>:

tigrawap:
Actually no need for lru_cache, it's already in memory by just importing
list from file... since file never imported by itself. And import levels
locks/memory management is as good as you can get with python...

2017-05-11 13:09 GMT+03:00 Anton Bykov <tigrawap@gmail.com>:

PR #20, opened 2017-05-09
warn usage of wait() or iter_wait() with predicate but no message

DoronCohen:


easypy/timing.py
```diff
@@ -201,7 +202,19 @@ def pred():
...
 
 def iter_wait(timeout, pred=None, sleep=0.5, message=None,
-              progressbar=True, throw=True, allow_interruption=False, caption=None):
+              progressbar=True, throw=True, allow_interruption=False, caption=None, from_wait=False):
+
+    # Calling wait() or iter_wait() with a predicate and no message is very
+    # not informative and is pending deprecation
+    if pred and not message:
+        func = 'wait' if from_wait else 'iter_wait'
+        stack_level = 3 if from_wait else 2
```
idanarye: And what about `repeat`, `wait_progress` and `iter_wait_progress`? This style can quickly get out of hand...

How about instead, we actually calculate the stack level?

```python
def stack_level_to_get_out_of_file():
    frame = currentframe().f_back
    filename = frame.f_code.co_filename
    stack_levels = 1
    while frame.f_code.co_filename == filename:
        stack_levels += 1
        frame = frame.f_back
    return stack_levels
```

This will ensure that no matter how deep the chain is, as long as it is contained in `wepy.timing` we'll get the correct stack level.

DoronCohen: That's great, thanks

DoronCohen: _There is no way to add a message to repeat, so it'll always print a warning._
If `throw` is False the no warning will be raised.\
@koreno

easypy/misc.py
```diff
@@ -35,3 +35,13 @@ def get_all_subclasses(cls):
                                    for subclass in cls.__subclasses__()
                                    for rec_subclass in get_all_subclasses(subclass)]
     return [subclass for subclass in _all if not hasattr(subclass, "_%s__is_mixin" % subclass.__name__)]
+
+
+def stack_level_to_get_out_of_file():
```
koreno: `get_levels_to_calling_module`

PR #21, opened 2017-05-10
Add `late_decorator` - run decorator in objects instead of classes

idanarye:
I've encountered several situations in the past where this decorator could be useful. The most recent one is @navotsil's big cluster PR. With this, `CACHE_TIMESTAMPS` and friends can be bundled in a class that will be part of `WekaSystem`, and we can use late decorators to easily use the methods of that class as decorators for `WekaSystem` properties.

### DIFF
--- a/easypy/decorations.py
+++ b/easypy/decorations.py
@@ -5,6 +5,7 @@ from functools import wraps, partial
 from types import MethodType
 import warnings
 from threading import RLock
+from operator import attrgetter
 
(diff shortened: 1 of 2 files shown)


koreno:
Let's ingest the big-clusters stuff first, and later see how it gets improved by this feature.
Currently it feels a bit niche for easypy...

koreno:
out of scope...

PR #23, opened 2017-05-18
Throttled: Should be decorator

tigrawap:


easypy/concurrency.py
```diff
@@ -1052,22 +1052,23 @@ def __call__(self):
...
+        def throttled_func(*args, **kwargs):
+            nonlocal last_run
+            ts = time.time() * 1000
+            run = False
+            with lock:
+                if ts - last_run > ms:
+                    last_run = ts
+                    run = True
+            if run:
+                return func(*args, **kwargs)
```
koreno: assert the return value is `None`

koreno: or use timecache

koreno:
write a test for it

PR #25, opened 2017-05-18
bunch: make it rigid by defining KEYS

koreno:
we may use this for various configuration structs


idanarye:
My take on this(on top of yours): #26

PR #26, opened 2017-05-20
Rigid bunch with black magic

idanarye:
If you're gonna do something, do it with awesome syntax!


koreno:
Nice, but no...
The magic is too black, and the outcome is too unconventional (defining members w/o assignment)...

PR #31, opened 2017-06-13
Close #13 - better lockstep API

### LINKED ISSUE #13: Better lockstep API
@koreno - what's our policy regarding breaking changes in easypy?

[You wanted to add `process.execute()`](https://github.com/weka-io/easypy/pull/9#discussion_r112005932) to ease the execution without stepping into the process. We decided against it because it's not super-trivial(we can't just put a method on the wrapper - it won't get bound to the object) and because we want the non-stepping invocation to be a direct method of the object, so we'd still need a wrapper method - which kind of nullifies the benefit of `process.execute()`.

But what if we reverse it? Instead of:

```python
@lockstep
def my_process_lockstep():
    # ...

# step-by-step
with my_process_lockstep() as process:
    # ...

# just do it
my_process_lockstep.execute()
```

We'll do this:
```python
@lockstep
def my_process():
    # ...

# step-by-step
with my_process.lockstep() as process:
    # ...

# just do it
my_process()
```

We'll still need our own property descriptor, but we'll have the benefit of having the direct execution as the main API, which will save us the need for a wrapper method.

idanarye:
Assuming:
```python
class Foo:
    @lockstep
    def process(self):
        # ...

foo = Foo()
```

* `foo.process()` - run the entire process
* `with foo.process.lockstep() as process:` - step into the process
* `yield from foo.process.lockstep()` - for lockstep nesting

### DIFF
--- a/easypy/lockstep.py
+++ b/easypy/lockstep.py
@@ -1,5 +1,5 @@
 from contextlib import contextmanager
-from functools import wraps
+from functools import wraps, update_wrapper
 
 from .exceptions import TException
 
@@ -47,6 +47,10 @@ class _LockstepInvocation(object):
         for self._current_step in self._generator:
             pass
 
+    def __iter__(self):
+        yield from self._generator
+        self._current_step = 'finished'
+
     def _next_step_and_value(self):
         yield_result = next(self._generator)
         if isinstance(yield_result, tuple) and len(yield_result) == 2:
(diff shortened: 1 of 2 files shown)


idanarye:
Not 100% sure about the `yield from` API. Maybe now that we are explicitly running the lockstep, we should also give the `yield from` it's own API? Something like `yield from foo.process.delegate()`?

idanarye:
BTW: please let me merge this. I want to merge it and immediately update testlight - otherwise it'll break the installation.

koreno:
Shouldn't it be
```
with foo.process.lockstep() as process:
    yield from process
```
?

idanarye:
[Currently we allow yielding from the context manager itself.](https://github.com/weka-io/easypy/blob/c638b2a64f8c0f59622309407b6e82908f7a0c17/tests/test_lockstep.py#L125)

Yielding from the process object is easy to implement, but I generally prefer APIs that allow their user to avoid binding new names.

I would like to have a separate method for yielding the steps, because having it under `.lockstep()` forces me to use `_LockstepContextManagerWrapper` instead of a regular `contextmanager`, but I struggle to find a good name for it:

* `yield from foo.process.unwrap_lockstep()`?
* `yield from foo.process.nested_lockstep()`?
* `yield from foo.process.lockstep_steps()`?

idanarye:
?

PR #32, opened 2017-06-24
concurrency: LoggedRLock - logs when lock is waited, and supports expiration

koreno:


easypy/concurrency.py
```diff
@@ -955,6 +958,77 @@ def _get_my_ident():
...
+
+        log_timer = Timer(expiration=None if timeout < 0 else timeout)
+        while not log_timer.expired:
+            timeout = min(log_timer.remain or self._log_interval, self._log_interval)
+            if self._expiration_timer:
+                timeout = min(self._expiration_timer.remain, timeout)
+            if self._lock.acquire(blocking=True, timeout=timeout):
+                # set the timer after acquiring the lock!
+                self._expiration_timer = self._expiration_timer or Timer(expiration=expiration or self._default_expiration)
+                _logger.debug("%s - acquired", self)
```
noopti: let's log "acquired by {}" ...

easypy/concurrency.py
```diff
@@ -955,6 +958,77 @@ def _get_my_ident():
...
+        try:
+            owner = threading._active[owner].name
+        except KeyError:
+            pass
+        if owner:
+            return "<{}: owned by <{}>x{} for {}>".format(self._name, owner, count, self._expiration_timer.elapsed)
+        else:
+            return "<{}: unowned>".format(self._name)
+
+    def acquire(self, blocking=True, timeout=-1, expiration=None):
```
tigrawap: Why both blocking and timeout?  
Is not timeout=None == non blocking and timeout explicitly blocking?

easypy/concurrency.py
koreno: Why?...

easypy/concurrency.py
```diff
@@ -955,6 +958,77 @@ def _get_my_ident():
...
+    def __init__(self, name=None, log_interval=15, default_expiration=None):
+        self._lock = self.__class__.LockType()
+        self._name = name or '{}-{:X}'.format(self.LockType.__name__, id(self))
+        self._default_expiration = default_expiration
+        self._expiration_timer = None
+        self._log_interval = log_interval
+
+        # we want to support both the gevent and builtin lock
+        try:
+            self._lock._owner
```
idanarye: And PEP8 is OK with this?

easypy/concurrency.py
idanarye: `timeout=-1`? Why not `None`?

easypy/concurrency.py
```diff
@@ -955,6 +958,77 @@ def _get_my_ident():
...
+            return self._lock.acquire(blocking=False)
+
+        log_timer = Timer(expiration=None if timeout < 0 else timeout)
+        while not log_timer.expired:
+            timeout = min(log_timer.remain or self._log_interval, self._log_interval)
+            if self._expiration_timer:
+                timeout = min(self._expiration_timer.remain, timeout)
+            if self._lock.acquire(blocking=True, timeout=timeout):
+                # set the timer after acquiring the lock!
+                self._expiration_timer = self._expiration_timer or Timer(expiration=expiration or self._default_expiration)
```
idanarye: If this fancy for _create-expiration-timer-if-not-already-exists_?

easypy/concurrency.py
koreno: consistency with python's API

PR #33, opened 2017-07-03
concurrency: bug fix LoggedRLock use of lease timer

koreno:
Would appreciate your careful review, in light of WEKAPP-36360

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -13,4 +13,4 @@ if not issubclass(logging.Logger, easypy_logging.ContextLoggerMixin):
     logging.Logger.__bases__ = logging.Logger.__bases__ + (easypy_logging.ContextLoggerMixin,)
 
 
-logging.basicConfig(level=logging.DEBUG)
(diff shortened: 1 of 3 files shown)


tests/test_concurrency.py
```diff
@@ -137,3 +140,24 @@ def do_lock():
         pass
 
     assert sum(c == call("%s - waiting...", lock) for c in _logger.debug.call_args_list) > 3
+
+
+def _test_logged_lock_races():
```
koreno: This unfortunately did not reproduce the bug I caught

tests/conftest.py
```diff
@@ -13,4 +13,4 @@
     logging.Logger.__bases__ = logging.Logger.__bases__ + (easypy_logging.ContextLoggerMixin,)
 
 
-logging.basicConfig(level=logging.DEBUG)
+logging.basicConfig(level=logging.DEBUG, format='%(asctime)s|%(process)2s:%(threadName)-25s|%(name)-40s|%(levelname)-5s|%(funcName)-30s |%(message)s')
```
koreno: separated into another commit

PR #35, opened 2017-07-20
Allow filtering by nested attributes

idanarye:
Motivation - easier filtering without lambdas/helper methods. Mostly useful for filtering by capabilities. Examples:

* We have this line:
    ```python
    all_hosts = all_hosts.filtered(user_app=True)
    ```
    But it requires to declare:
    ```python
    @property
    def user_app(self):
        return self.capabilities.user_app.relevant
    ```
    Now we can just:
    ```python
    all_hosts = all_hosts.filtered(capabilities__user_app=True)
    # No special method required
    ```

* We have this line:
    ```python
    return self.progress_timestamp_for(self.system.hosts.filtered(lambda h: h.capabilities.events.works))
    ```
    Now we can just:
    ```python
    return self.progress_timestamp_for(self.system.hosts.filtered(capabilities__events__works=True))
    ```

### DIFF
--- a/easypy/collections.py
+++ b/easypy/collections.py
@@ -2,7 +2,7 @@ from __future__ import absolute_import
 import collections
 from numbers import Integral
 from itertools import chain, islice
-from functools import partial
+from functools import partial, reduce
 import random
 from .predicates import make_predicate
 from .tokens import UNIQUE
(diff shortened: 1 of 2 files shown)


koreno:
Sorry buddy, it crosses the magic line for me. At XIV we had a lambda generator that did this:
`collection.choose(SELECTED.some_val > 5)`, which was nice, but when people tried to combine with additional conditions it all fell apart in a confusing way.

easypy shouldn't have such naming-sensitive features

PR #37, opened 2017-08-13
added `.zip_with` to MultiObject

koreno:


easypy/concurrency.py
```diff
@@ -599,6 +599,10 @@ def chain(self):
         "Chain the iterables contained by this MultiObject"
         return self.__class__(chain(*self))
 
+    def zip_with(self, *collections):
+        assert all(len(coll) == len(self) for coll in collections), "All collection must be of length %s" % len(self)
```
idanarye: Is this necessary? This makes it hard to use with infinite generators like `itertools.cycle`...

idanarye:
Is the tokens commit supposed to be here?

koreno:
Yeah, unrelated, but not worth It's own pr

easypy/concurrency.py
koreno: I think that's dangerous... at the very least `self.items` should all be paired, to avoid subtle mistakes

koreno: It's the type of thing I could discard when I encounter an real use case

idanarye: How about this then:

```python
def zip_with(self, *collections):
    result = self.__class__(zip(self, *collections))
    assert len(self) == len(result), "All collection must be of at least length %s" % len(self)
    return result
```

koreno:
Clearly that's better
😒

PR #39, opened 2017-08-23
Close #38: Add `TypedStruct`

### LINKED ISSUE #38: TypedStruct - schematic classes for structured people
We want a schema for our configuration - a declarative way to define what fields it should have and all sorts of info on these fields. We also want to be able to reflectively use these schemas - create Plumbum CLI flags, store the info in databases, send it to external services(like AWS) etc.

Goals:
* Allow creating types with lists of fields.
    * Automatic c'tor for these fields.
    * Getters and setters should work.
* Allow declaring the types of the fields.
    * Types should be verified at runtime, in the c'tor and in the setters.
* Allow declaring constraints on the fields(e.g. - "number can be from 13 to 42")
    * Of course, constraints should also be verified at runtime.
* Allow custom metadata - e.g. help text for Plumbum CLI.
* Allow default value for fields. Fields without default values are mandatory.
* Allow nesting of types.
* Should work well with IDEs' autocompletion.
* Nice-to-have: integration with MyPy. Not sure if it's even possible though - certainly not without Python 3.6's variable annotation...

idanarye:
* Simple definition syntax - without repeating field names.
* Short syntax for simple fields - workds with IDE completion.
* Full `ts.Field(...)` syntax to define default value, validators etc.
* `meta` field that can be used to provide metadata for reflection.
* Support for typechecked `list`, `dict` and `Bunch` fields.

Future plans(will need more design):

* Hooks/signals for doing things when users change the fields.
* Bubble up? Parent `TypedStruct`'s hook get activated when
  fields of nested `TypedStruct`s are changed?

### DIFF
--- a/easypy/typed_struct.py
+++ b/easypy/typed_struct.py
@@ -1,0 +1,407 @@
+from copy import deepcopy
+
+from .exceptions import TException
+from .tokens import MANDATORY
+from .collections import ListCollection, PythonOrderedDict
+from .bunch import Bunch, bunchify
+
+
+class InvalidFieldType(TException):
+    template = 'Bad field type {field_type} - {reason}'
+
+
+class NotFields(TException, AttributeError):
+    template = '{field_names} are not fields of {typed_struct.__name__}'
+
+
+class NotAField(TException, AttributeError):
+    template = '{field_name} is not a field of {typed_struct.__name__}'
+
+
+class NotAKey(TException, KeyError):
+    template = '{field_name} is not a key of {typed_struct.__name__}'
+
+
+class MissingField(TException, TypeError):
+    template = '{typed_struct.__name__} is missing mandatory field {field.name}'
+
+
+class FieldTypeMismatch(TException, TypeError):
+    template = '{field.name} is {expected_type}, not {received_type}'
+
+
+class FieldKeyTypeMismatch(TException, TypeError):
+    template = 'keys of {field.name} are {expected_type}, not {received_type}'
+
+
+class Field(object):
+    """
+    Define a single field in a TypedStruct
+
+        type: The type of the field
+              * Use [<type>] for a typechecked list
+              * Use {<key_type>: <type>} for a typechecked dict
+              * Use {str: <type>} for a typechecked Bunch
+        default: Default value for the field
+        preprocess: A function for preprocessing the values assigned to the field:
+                    * If the field can be converted from different types,
+                      should do the conversion and return the converted value.
+                    * If there are validation checks, should perform them and
+                      raise an exception if they fail.
+                    * Use the field.add_validation() and field.add_conversion()
+                      helpers to modify the preprocess function without
+                      manually setting this parameter.
+                    * Must return a value of the field's type.
+        meta: Custom metadata. Can be queried by tools that use reflection on
+              the TypedStruct object.
+    """
+
+    def __init__(self, type, *, default=MANDATORY, preprocess=None, meta=Bunch()):
+        if isinstance(type, list):
+            if len(type) != 1:
+                raise InvalidFieldType(field_type=type,
+                                       reason='Collection-style fields must be defined with a single item collection')
+            type, = type
+            self.collection_type = TypedList
+            self.key_type = None
+        elif isinstance(type, dict):
+            if len(type) != 1:
+                raise InvalidFieldType(field_type=type,
+                                       reason='Collection-style fields must be defined with a single item collection')
+            (self.key_type, type), = type.items()
+            if self.key_type is str:
+                self.collection_type = TypedBunch
+            else:
+                self.collection_type = TypedDict
+        else:
+            self.collection_type = None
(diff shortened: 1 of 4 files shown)


easypy/typed_struct.py
```diff
@@ -0,0 +1,366 @@
...
+
+
+class Field(object):
+    """
+    Define a single field in a TypedStruct
+
+        type: The type of the field
+              * Use [type] for a typechecked list
+              * Use {key_type: type} for a typechecked dict
+              * Use {str: type} for a typechecked Bunch
```
koreno: make me thing you want me to put _a_ str, and not specifically `str`...
maybe:
```
              * Use [<type>] for a typechecked list
              * Use {<key_type>: <type>} for a typechecked dict
              * Use {str: <type>} for a typechecked Bunch
```

easypy/typed_struct.py
```diff
@@ -0,0 +1,366 @@
...
+    Define a single field in a TypedStruct
+
+        type: The type of the field
+              * Use [<type>] for a typechecked list
+              * Use {<key_type>: <type>} for a typechecked dict
+              * Use {str: <type>} for a typechecked Bunch
+        default: Default value for the field
+        validate: Validator predicate(s) that will be called on the value.
+                  Should return False if it's no OK.
+        convert: Dictionary of converters from source type to the field's type.
```
koreno: in plumbum the validator is also a converter (returns the normalized value) - couldn't we do that here too?

easypy/typed_struct.py
```diff
@@ -0,0 +1,366 @@
...
+        self.default = default
+        self.validate = listify(validate)
+        self.convert = dict(convert)  # make sure it's a different dict
+        self.meta = Bunch(meta)
+        self.name = None
+
+        if issubclass(self.type, TypedStruct):
+            if self.default is MANDATORY:
+                try:
+                    default = self.type()
```
idanarye: Yea... and apparently my tests did not cover this one...

idanarye:
@koreno: I've merged `validate` and `convert` into `preprocess`, but also added helper methods to `Field` for setting them individually.

Old style:
```python
class Foo(ts.TypedStruct):
    a = ts.Field(int,
                 # Only even numbers between 10 and 30
                 validate=[lambda n: n % 2 == 0, range(10, 31).__contains__],
                 # Accept strings and floats by converting them to int
                 convert={str: int, float: int})
```

New style - manually implemented function:
```python
def preprocess(obj):
    if isinstance(obj, str) or isinstance(obj, float):
        obj = int(obj)

    assert obj % 2 == 0, 'not an even number'
    assert obj in range(10, 31), 'out of range([10, 30])'

    return obj

class Foo(ts.TypedStruct):
    a = ts.Field(int, preprocess=preprocess)
```

New style - helper functions:

```python
class Foo(ts.TypedStruct):
    a = ts.Field(int)

    a.add_validation(lambda n: n % 2 == 0, Exception, 'not an even number')
    a.add_validation(range(10, 31).__contains__, Exception, 'out of range([10, 30])')

    a.add_conversion(str, int)
    a.add_conversion(float, int)
```

The helper functions style is preferred, because with our manual implementation the first validation(number is even) will fail with a generic `TypeError` if if gets a non-numeric value, instead of failing on the more specific `InvalidFieldType`. We can't check the type before, because then `preprocess` won't be able to do conversions.

easypy/typed_struct.py
```diff
@@ -0,0 +1,407 @@
...
+              * Use [<type>] for a typechecked list
+              * Use {<key_type>: <type>} for a typechecked dict
+              * Use {str: <type>} for a typechecked Bunch
+        default: Default value for the field
+        preprocess: A function for preprocessing the values assigned to the field:
+                    * If the field can be converted from different types,
+                      should do the conversion and return the converted value.
+                    * If there are validation checks, should perform them and
+                      raise an exception if they fail.
+                    * Use the field.add_validation() and field.add_conversion()
```
koreno: Excellent!

koreno: though... now that I see the api and implementation (with predicate, ex_type, etc.) I think it's quite an overkill and "YAGNI"... But if you insist...

What does @tigrawap  think?

tests/test_typed_struct.py
```diff
@@ -0,0 +1,439 @@
...
+
+
+@pytest.mark.parametrize("use_full_syntax", [True, False])
+def test_typed_struct_dict_fields(use_full_syntax):
+    class Foo(ts.TypedStruct):
+        a = int
+
+    if use_full_syntax:
+        class Bar(ts.TypedStruct):
+            foos = ts.Field({int: Foo})
```
koreno: just so I'm clear - if we could we'd do `ts.Field(Dict[int, Foo])`, right?

idanarye: Right. And `ts.Field(Bunch[Foo])` instead of `ts.Field({str: Foo})` too.

PR #40, opened 2017-08-28
Tokens fix

idanarye:


easypy/misc.py
```diff
@@ -18,12 +18,32 @@ def __repr__(self):
 
 class Token(str):
 
+    _all = {}
+
     def __new__(cls, name):
```
DoronCohen: This doesn't look thread safe if it matters

PR #41, opened 2017-08-29
PersistentCache: Support optional default

tigrawap:


easypy/caching.py
```diff
@@ -22,6 +22,7 @@
 _disable_persistance = bool(os.getenv("bamboo_planKey", os.getenv("WEKA_job_key", None)))  # disable this feature when running from bamboo
 
 
+_NONE = object()
```
koreno: use an easpy token, that's what they're for.
`from easypy.tokens import NO_DEFAULT`

unless there's some circular import... but there shouldn't be...

tigrawap: can't agree with this :/

One thing when you really need to pass token between parts of app - than well, maybe.
Another thing is when it's such case.... Absolutely unnecessary overhead and complexity. 

"Use it since we have it" - this is the only reason and it's not good enough

Why would I want to worry, that someone passes this token as default, due to some logic in app, when this parameter was passed from some function signature, where default was NO_DEFAULT?

Issue #44, opened 2017-08-30
singleton_context_manager does not play well with RLock


idanarye:
```python
from easypy.decorations import singleton_contextmanager_method
from threading import RLock, Thread, Event
from time import sleep


def test_singleton_contextmanager_method_with_rlocks_bad_order():
    class Foo:
        lock = RLock()

        @singleton_contextmanager_method
        def cm(self):
            with self.lock:
                yield

    foo = Foo()
    event1 = Event()
    event2 = Event()

    def lock_and_wait(event):
        with foo.cm():
            event.wait()

    thread1 = Thread(target=lock_and_wait, args=(event1,))
    thread1.start()
    sleep(1)
    thread2 = Thread(target=lock_and_wait, args=(event2,))
    thread2.start()

    event1.set()
    thread1.join()

    event2.set()
    thread2.join()
```

Running this test prints:

```
$ py.test tests/ -s -k test_singleton_contextmanager_method_with_rlocks_bad_order
========================================== test session starts ===========================================
platform linux -- Python 3.6.2, pytest-3.0.6, py-1.4.32, pluggy-0.4.0
rootdir: /files/code/easypy, inifile: pytest.ini
collected 118 items

tests/test_decorations.py Exception in thread Thread-3:
Traceback (most recent call last):
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/files/code/easypy/tests/test_decorations.py", line 87, in lock_and_wait
    event.wait()
  File "/usr/lib/python3.6/contextlib.py", line 88, in __exit__
    next(self.gen)
  File "/files/code/easypy/easypy/decorations.py", line 114, in inner
    yield val
  File "/files/code/easypy/easypy/decorations.py", line 90, in __exit__
    self.ctm.__exit__(*sys.exc_info())
  File "/usr/lib/python3.6/contextlib.py", line 88, in __exit__
    next(self.gen)
  File "/files/code/easypy/tests/test_decorations.py", line 79, in cm
    yield
RuntimeError: cannot release un-acquired lock

.

========================================== 117 tests deselected ==========================================
================================ 1 passed, 117 deselected in 1.31 seconds ================================
```

(the test still passes, but there is an exception in a thread)

[...]

PR #45, opened 2017-08-31
Add TeamsLock

idanarye:
* Create a lock with a list of teams.
* When entering the lock you must declare your team.
* If the lock is acquired by your team you may enter.
* If the lock is acquired by a different team you will block.

### DIFF
--- a/easypy/concurrency.py
+++ b/easypy/concurrency.py
@@ -1240,3 +1240,78 @@ class synchronized_on_first_call():
             if not self.initialized:
                 self.initialized = True
             return ret
+
+
+class TeamsLock(object):
+    """
+    The lock always belong to one "team"(or no team), and other team members
+    can join the lock - but not members of other teams!
+
+    * Initialize with a list of team names
+    * Use the `.acquire_for_team(<team-name>)` context manager to lock for your team
+
+    Example:
+
+        lock = TeamsLock('foo', 'bar')
+
+        with lock.acquire_for_team('foo'):
+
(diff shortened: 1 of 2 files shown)


idanarye:
This is something I wanted to insert a while ago, but @koreno said my use-case is not good enough. Well - now it is!

When we start/stop Stress0 we need to pause the Mayhem - S0 can handle failures while it runs but not during the start/stop sequences. We used to have only one thing starting/stopping S0(the test), so `RWLock` was good enough - the mayhem actions were the readers and the test was the writer and multiple actions could run together but not when the test is trying to do things.

Now that we have filesystem actions, this is not enough:
* When we remove a filesystem group, we need to remove it's filesystems first
* When we remove a filesystem, we need to stop stress0 on it
* Stopping S0 takes some time, so we want to parallelize the process
* But... if every _stop-S0-on-filesystem_ thread will try to pause the mayhem in parallel - they'll just block each other(multiple writers)

As a poor man's `TeamsLock` I've strapped a `singleton_contextmanager_method` on the mayhem pause - but this does not work very well(#44). So I want to add `TeamsLock` as a solution to this problem.

P.S.: This _can_ be solved by locking **before** starting the concurrent subtasks, but we have four levels(entire system(during cleanup), filesystem group, filesystem and snapshot), and each must be aware of whether or not it is the top level. This can become quite a mess...

koreno:
So which teams would we have in our Mayhem/Stress0 scenario?

PR #53, opened 2017-11-23
Properly display timezone in `short_utc_fmt`

idanarye:
- UTC timezone will be displayed as 'Z'.
- Named timezone will be displayed with their name(e.g. 'IST')
- Nameless timezones will be displayed with their offset(e.g. '+02:00')

### DIFF
--- a/easypy/timezone.py
+++ b/easypy/timezone.py
@@ -98,6 +98,10 @@ def parse_isoformat(source):
         return timestamp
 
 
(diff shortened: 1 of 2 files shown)


idanarye:
We need this for https://wekaio.atlassian.net/browse/WEKAPP-50186

koreno:
Aren't there other packages to deal with all this? `dateutils`?... feels like easypy isn't the ultimate place for all this...

PR #55, opened 2017-12-06
Let the predicate of `wait` know if it's the final attempt

idanarye:
Use case: on the final attempt, if the predicate is still `False`, it
may want to print something to the log to help us determine what went
wrong. To avoid spamming the logs, we don't want to print it on every
iteration.

### DIFF
--- a/easypy/timing.py
+++ b/easypy/timing.py
@@ -12,7 +12,7 @@ from .units import Duration
 from .exceptions import PException
 from .humanize import time_duration  # due to interference with jrpc
 from .misc import stack_level_to_get_out_of_file
-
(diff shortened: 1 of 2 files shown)


idanarye:
@koreno - I know you have [reverted a similar looking feature on `retry`](https://github.com/weka-io/easypy/commit/2e55cef966a2712e70c1474068db092ba7f46e86), but in this case the predicate was not expecting any arguments to begin with, so there is no risk of messing with the user's arguments.

idanarye:
Actual use case: https://wekaio.atlassian.net/browse/WEKAPP-51846

idanarye:
BUMP

PR #56, opened 2017-12-26
Add `ASSERT` for easily creating asserion error messages

idanarye:
I talked about fluent assertions, but I think I sketched up something even better. The idea: write regular Python expression, but mark certain parts of it to be replaced with human language text.


koreno:
The magic is too great with this one...
Doesn't belong in easypy imho.

PR #57, opened 2018-01-18
Verify collection type in `TypedStruct`

idanarye:
To prevent things like this:

```python
In [1]: from easypy import typed_struct as ts

In [2]: class Foo(ts.TypedStruct):
   ...:     bars = [str]
   ...:

In [3]: Foo(bars='hello')
Out[3]:
Foo(bars=['h',
          'e',
          'l',
          'l',
          'o'])
```

### DIFF
--- a/tests/test_typed_struct.py
+++ b/tests/test_typed_struct.py
@@ -453,3 +453,52 @@ def test_typed_struct_preprocess(use_helper_methods):
 
     with pytest.raises(Error2):
         Foo(a='32')
+
+
+def test_typed_struct_collection_type_verification():
+    """
+    To prevent things like this:
+
+        In [1]: from easypy import typed_struct as ts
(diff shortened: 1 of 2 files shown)


tests/test_typed_struct.py
```diff
@@ -453,3 +453,52 @@ class Foo(ts.TypedStruct):
...
+
+    # Test TypedDict:
+
+    assert Foo(b={'5': '6'}).b == {'5': '6'}
+    with pytest.raises(ts.FieldCollectionTypeMismatch):
+        Foo(b='8')
+    with pytest.raises(ts.FieldCollectionTypeMismatch):
+        Foo(b=['9'])
+
+    # Test TypedBunch:
```
koreno: Bunch?

PR #59, opened 2018-02-01
Concurrency synchronization coordinator

idanarye:
@koreno - before I start testing it with testlight (which will increase the cycles) please approve the design.

### DIFF
--- a/easypy/concurrency.py
+++ b/easypy/concurrency.py
@@ -366,6 +366,20 @@ def async(func, params=None, workers=None, log_contexts=None, final_timeout=2.0,
 
     funcname = _get_func_name(func)
 
+    try:
+        signature = inspect.signature(func)
+    except ValueError:
+        # In Python 3.5+, inspect.signature returns this for built-in types
+        pass
+    else:
+        if '_sync' in signature.parameters and '_sync' not in kw:
+            assert len(params) <= executor._max_workers, 'SynchronizationCoordinator with %s tasks but only %s workers' % (
+                len(params), executor._max_workers)
+            synchronization_coordinator = SynchronizationCoordinator(len(params))
+            kw['_sync'] = synchronization_coordinator
+
+            func = synchronization_coordinator._abandon_when_done(func)
+
     futures = Futures()
     for args, ctx in zip(params, log_contexts):
         future = executor.submit(_run_with_exception_logging, func, args, kw, ctx)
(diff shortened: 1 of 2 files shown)


easypy/concurrency.py
```diff
@@ -1268,3 +1281,165 @@ def __call__(cls, *args, **kwargs):
...
+    argument, it will replace it with a proper SynchronizationCoordinator instance:
+
+        def foo(a, _sync=SYNC):
+            _sync.wait_for_everyone()
+            _sync.collect_and_call_once(a, lambda a_values: print(a))
+
+        MultiObject(range(5)).call(foo)
+    """
+
+    def __init__(self, num_participators):
```
koreno: "participant"

easypy/concurrency.py
```diff
@@ -1268,3 +1281,165 @@ def __call__(cls, *args, **kwargs):
...
+
+    def _reset_barrier(self):
+        self.barrier = threading.Barrier(self.num_participators, action=self._post_barrier_action)
+
+    def _post_barrier_action(self):
+        if self.num_participators != self.barrier.parties:
+            self._reset_barrier()
+        self._wait_context = None
+
+        if self._call_once_function:
```
koreno: Consider:
```
call_once_function, self._call_once_function = self._call_once_function, None
if call_once_function:
    ...
```

easypy/concurrency.py
```diff
@@ -1268,3 +1281,165 @@ def __call__(cls, *args, **kwargs):
...
+            self._call_once_collected_param = []
+
+            try:
+                self._call_once_result = call_once_function(collected_param)
+            except BaseException as e:
+                self._reset_barrier()
+                self._call_once_result = e
+                raise
+
+    def wait_for_everyone(self):
```
koreno: There should be a `timeout` here I think, in case some rogue thread takes too long.
An TimeoutException raised here would rise out of all of the waiting threads up to `async`, which should somehow know to abandon the lagging thread (and allow it to crash (?) on the broken barrier if it ever reaches it finally)

idanarye:
I just realized that I neglected the shared lock use-case. I need to add a way to run a context manager in one thread and make sure the `__exit__` is executed in the same thread that executed the `__enter__`.

easypy/concurrency.py
idanarye: Adding a timeout is easy - making it abandon the threads that lag behind will be harder (unless I split the coordinator, or enforce usage of thread identities). I'd rather just invalidate the entire coordinator.

PR #61, opened 2018-02-16
Caching with key func

koreno:


easypy/caching.py
```diff
@@ -171,29 +172,73 @@ def clear():
...
+                elif param.kind is _VAR_POSITIONAL:
+                    val = ()
+                elif param.kind is _VAR_KEYWORD:
+                    val = {}
+                else:
+                    # This BoundArguments was likely produced by
+                    # Signature.bind_partial().
+                    continue
+                new_arguments.append((name, val))
+        bound_arguments.arguments = OrderedDict(new_arguments)
```
idanarye: Why convert a list of pairs to an `OrderedDict`? Why not just make an empty `OrderedDict` and populate it?

easypy/caching.py
```diff
@@ -171,29 +172,73 @@ def clear():
...
+else:
+    def _apply_defaults(bound_arguments):
+        return bound_arguments.apply_defaults()
```
idanarye: `bound_arguments.apply_defaults()` returns `None`, but it's still weird that this version "returns" something while the backward-compatibility version doesn't.

easypy/caching.py
koreno: I copied the code from python3.5 sources, didn't want to change anything to official code.

easypy/caching.py
koreno: I'm like a decorator, don't care about implementation detail, I want to be future-compatible.

idanarye: Future-compatibility-wise, if in the future apply_default will return something, someone might use the value returned from _apply_defaults believing it's backward compatible

tests/test_misc.py
```diff
@@ -7,3 +7,14 @@ def foo(p=AUTO):
...
+    assert MAX == "MAX"
+    assert MAX == "<MAX>"
+    assert MAX == "max"
+    assert MAX == "<max>"
+
+    d = {AUTO: AUTO, MAX: MAX}
+    assert d[AUTO] == AUTO
+    assert d[MAX] == MAX
+    assert d['<MAX>'] is MAX
+    assert 'AUTO' not in d
```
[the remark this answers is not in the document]
koreno: I'll revisit this at some point...

koreno:
(already merged)

PR #64, opened 2018-03-22
Allow limiting the omitable params in `kwargs_resilient`

idanarye:


koreno:
"omitable"! why didn't you propose that earlier?!... :)

idanarye:
It's not too late? The only thing that uses it is `iter_wait` - it's still easy to change the name.

PR #65, opened 2018-03-27
Add logged condition

idanarye:
Note that unlike `LoggedRLock`, no one is owning the condition while the waiters wait for it - so there is no point in logging the "number of owners".

### DIFF
--- a/easypy/concurrency.py
+++ b/easypy/concurrency.py
@@ -26,7 +26,7 @@ from easypy.humanize import IndentableTextBuffer, time_duration
 from easypy.misc import Hex
 from easypy.humanize import format_thread_stack
 from easypy.threadtree import iter_thread_frames
-from easypy.timing import Timer
+from easypy.timing import Timer, TimeoutException
 from easypy.units import NEVER, MINUTE, HOUR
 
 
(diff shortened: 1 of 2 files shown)


easypy/concurrency.py
```diff
@@ -1484,3 +1484,98 @@ def collect_and_call_once(self, param, func):
...
+            _logger.debug('%s - waiting to be acquired for ' + msg, self, *args)
+        try:
+            yield
+        finally:
+            self._cond.release()
+
+    @contextmanager
+    def notifying_all(self, msg, *args):
+        """
+        Acquire the condition lock for the context, and notify all all waiters afterward.
```
koreno: all all

easypy/concurrency.py
```diff
@@ -1484,3 +1484,98 @@ def collect_and_call_once(self, param, func):
...
+        Users should run the command that triggers the conditions inside this context manager.
+        """
+        with self._acquired_for('performing a %s notifying all waiters' % msg, *args):
+            yield
+            _logger.debug('%s - performed: ' + msg, self, *args)
+            self._cond.notifyAll()
+
+    def wait_for(self, pred, msg, *args, timeout=None):
+        """
+        Wait for a a predicate. Only check it when notified.
```
koreno: `a a`?...

tests/test_concurrency.py
```diff
@@ -257,3 +259,75 @@ def test_multiexception_api():
...
+    progress = 0
+    executed = []
+
+    def wait_for_progress_to(progress_to):
+        cond.wait_for(lambda: progress_to <= progress, 'progress to %s', progress_to)
+        executed.append(progress_to)
+
+    with concurrent(wait_for_progress_to, 10), concurrent(wait_for_progress_to, 20), concurrent(wait_for_progress_to, 30):
+        with patch("easypy.concurrency._logger") as _logger:
+            sleep(1)
```
koreno: (i try to make the sleeps as small as reasonably possible os that running pytest is quick)

tests/test_concurrency.py
```diff
@@ -257,3 +259,75 @@ def test_multiexception_api():
...
+
+        assert sum(c == call("%s - waiting for progress to %s", cond, 10) for c in _logger.debug.call_args_list) > 3
+        assert sum(c == call("%s - waiting for progress to %s", cond, 20) for c in _logger.debug.call_args_list) > 3
+        assert sum(c == call("%s - waiting for progress to %s", cond, 30) for c in _logger.debug.call_args_list) > 3
+        assert executed == []
+
+        with patch("easypy.concurrency._logger") as _logger:
```
koreno: why not patch it for the entire test?

easypy/concurrency.py
idanarye: So nice I wrote it twice

tests/test_concurrency.py
idanarye: Because I run different checks on each one I patch, so it's easier to not have the calls mixed.

PR #67, opened 2018-04-11
concurrency: support easier MultiException handling by auto-generating subtypes

koreno:
This is cool!


idanarye:
This fails:
```python
    class Foo(Exception):
        pass

    class Bar(Foo):
        pass

    def raise_it(exception_class):
        raise exception_class

    with pytest.raises(MultiException.Foo):
        MultiObject([Bar]).call(raise_it)
```

This works:
```python
    class Namespace:
        class Foo(Exception):
            pass

    class Foo(Exception):
        pass

    def raise_it(exception_class):
        raise exception_class

    with pytest.raises(MultiException.Foo):  # which `Foo` am I referring to?
        MultiObject([Namespace.Foo]).call(raise_it)
```

I don't like this matching-names based API. I'd rather have something like `MultiException[Foo]` instead:

* It works with namespaces - Python knows exactly which `Foo` you are talking about.
* It will be possible to properly support subclassing.
* It fits the (newish) convention for "generics" in Python (see https://docs.python.org/3/library/typing.html)

PR #68, opened 2018-04-15
concurrency: fix multiexception rendering to include timestamp and context; add timestamp exceptions if concurrent

koreno:
There's a bunch of code duplication here, but it's too much to refactor at this stage. In other words, we're looking for bugs, not so much for code elegance...

### DIFF
--- a/easypy/concurrency.py
+++ b/easypy/concurrency.py
@@ -18,9 +18,10 @@ import signal
 import threading
 import time
 import warnings
+from datetime import datetime
 
 from easypy.decorations import parametrizeable_decorator
-from easypy.exceptions import TException, PException
+from easypy.exceptions import TException, PException, make_block
 from easypy.gevent import is_module_patched, non_gevent_sleep, defer_to_thread
 from easypy.humanize import IndentableTextBuffer, time_duration
 from easypy.misc import Hex
@@ -28,6 +29,7 @@ from easypy.humanize import format_thread_stack
 from easypy.threadtree import iter_thread_frames
 from easypy.timing import Timer, TimeoutException
 from easypy.units import NEVER, MINUTE, HOUR
+from easypy.colors import colorize_by_patterns
 
 
 this_module = import_module(__name__)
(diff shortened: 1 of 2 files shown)


easypy/concurrency.py
```diff
@@ -225,29 +240,31 @@ def _get_buffer(self, **kw):
...
+                ts = datetime.fromtimestamp(exc.timestamp).isoformat()
+                buff.write(normalize_color("MAGENTA<<Timestamp: %s>>" % ts))
+            if kw.get("context", True) and exc.context:
+                buff.write("Context: %s" % _format_context(exc.context))
+
+        add_details(self)
+
         for exc in self.actual:
             with buff.indent("{.__class__.__qualname__}", exc):
-                if isinstance(exc, self.__class__):
```
koreno: (This is actually a bug now that MultiException subtyping is in...)

easypy/concurrency.py
```diff
@@ -343,15 +360,16 @@ def logged_wait(self, timeout=None, initial_log_interval=None):
 
 
 def _run_with_exception_logging(func, args, kwargs, ctx):
+    ctx.update(threadname=threading.current_thread().name)
```
idanarye:
```python
ctx['threadname'] = threading.current_thread().name
```

(yea, I know it's unrefactored code, but there is no reason not to use the conventional style here...)

easypy/concurrency.py
```diff
@@ -936,6 +954,10 @@ def _logged_func(self):
         except Exception as exc:
             _logger.silent_exception("Exception in thread running %s (traceback can be found in debug-level logs)", self.func)
             self.exc = exc
+            try:
+                exc.timestamp = time.time()
+            except Exception:
```
idanarye: What kind of `Exception` are we expecting here?

koreno: if exc happens to have __slots__ or some __setattribute__ or a timestamp readonly property it might not allow arbitrary attribute assignment.

idanarye: And we don't want a more specific exception? Or we just don't want to take the chance an exception logging related code will fail?

PR #72, opened 2018-04-23
Automatically convert fields in `TypedStruct` definition to `Field`

idanarye:
Previously, if you wanted to - for example - add a conversion to a
field, you'd have to do this:

```python
class Foo(ts.TypedStruct):
    a = ts.Field(int)
    a.convertible_from(str)
```

Or this:

```python
class Foo(ts.TypedStruct):
    a = int
Foo.a.convertible_from(str)
```

With this commit, you can do this:

```python
class Foo(ts.TypedStruct):
    a = int
    a.convertible_from(str)
```

This is the first step of a better `TypedStruct` API that I'm planning. For the next step, I want to support things like this:

```python
class Foo(TypedStruct):
    with convertible_from(int, float, str):
        a = int
        b = float
        c = [int]
```

Note that currently we do things like this:

```python
def numeric_field(field_type, **kwargs):
    field = ts.Field(field_type, **kwargs)
    field.add_conversion(int, float, str)
    return field

class Foo(ts.TypedStruct):
    a = numeric_field(int)
    b = numeric_field(float)
    c = numeric_field([int])
```

### DIFF
--- a/easypy/typed_struct.py
+++ b/easypy/typed_struct.py
@@ -1,7 +1,7 @@
 from copy import deepcopy
 
 from .exceptions import TException
-from .tokens import MANDATORY
+from .tokens import AUTO, MANDATORY
 from .collections import ListCollection, PythonOrderedDict, iterable
 from .bunch import Bunch, bunchify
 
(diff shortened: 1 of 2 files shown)


idanarye:
I also consider supporting this:
```python
class Foo(ts.TypedStruct):
    a = int
    _.convertible_from(str)
```

Where `_` acts like in IPython and refers to the last attribute (if it's a field - I want to make it an error if used after declaring a constant or a method)

This should prevent the need to repeat the field name, making it more copy-paste friendly.

idanarye:
BUMP?

DoronCohen:
sorry for not reviewing. I feel a little out of the loop with the `TypedStruct` thing. anyway this looks like an improvement to it's API to me

PR #73, opened 2018-04-29
concurrency: improve logging when waiting on threads

koreno:
- use thread name instead of thread-ident
- fix lookup of frame when waiting on Futures
- find a useful frame, non-threading-related when reporting hanging Futures


idanarye:
I'd still like to have the thread's name, to be able to search for it `teka threads`...

PR #74, opened 2018-04-30
timing: fix wait to not swallow the 'PredicateNotSatisfied' of a nest…

koreno:
…ed wait call


easypy/timing.py
```diff
@@ -286,6 +286,9 @@ def sleep():
             try:
                 ret = pred(is_final_attempt=bool(expired))
             except PredicateNotSatisfied as _exc:
+                if getattr(_exc, "duration", 0):
```
idanarye: What if the user sets an attribute named `duration`? Shouldn't we add an attribute - like `_already_timed_out`?

PR #76, opened 2018-05-08
Define typed struct fields with context managers

idanarye:
The magic here is so strong you can start calling me "Idandalf the White", but this will allow much more elegant definition of typed structs.

For example, we have something like this in our mayhem configuration:

```python
    def numeric_field(field_type, **kwargs):
        field = ts.Field(field_type, **kwargs)
        field.add_validation(lambda n: 0 <= n, lambda: NegativeNumbersInMayhemConfig(field_name=field.name))
        for convert_from in [int, float, str]:
            if convert_from is not field.type:
                field.add_conversion(convert_from, field.type)
        return field


    class MayhemConfig(ts.TypedStruct):
        errors = numeric_field({str: float})
        sleeping_intervals = numeric_field([Duration])
        extra_action_sleeping_intervals = numeric_field([Duration])
        protection_from_removal_period = numeric_field([Duration])
        rebuild_while_down_probability = numeric_field(float, default=0.0)
        extra_action_period = numeric_field(Duration, default=0)
        shrink_default_filesystem = numeric_field(float, default=0.0)
        wait_for_all_actions_end_probablility = numeric_field(float, default=0.0)
```

With this new feature, we could make it look like this:

```python
    def numeric_field(field):
        field.add_validation(lambda n: 0 <= n, lambda: NegativeNumbersInMayhemConfig(field_name=field.name))
        for convert_from in [int, float, str]:
            if convert_from is not field.type:
                field.add_conversion(convert_from, field.type)


    class MayhemConfig(ts.TypedStruct):
        with FIELD_SETTING(numeric_field):
            errors = {str: float}
            sleeping_intervals = [Duration]
            extra_action_sleeping_intervals = [Duration]
            protection_from_removal_period = [Duration]
            with DEFAULT(0):
                rebuild_while_down_probability = float
                extra_action_period = Duration
                shrink_default_filesystem = float
                wait_for_all_actions_end_probablility = float
```

### DIFF
--- a/easypy/collections.py
+++ b/easypy/collections.py
@@ -1,9 +1,10 @@
 from __future__ import absolute_import
 import collections
 from numbers import Integral
-from itertools import chain, islice
+from itertools import chain, islice, count
 from functools import partial
 import random
+from contextlib import contextmanager
 from .predicates import make_predicate
 from .tokens import UNIQUE
 
@@ -684,3 +685,34 @@ class SlidingWindow(list):
         super().append(item)
         if len(self) > self.size:
             self.pop(0)
+
+
+class ContextCollection(object):
+    """
+    A collection where you add and remove things using context managers::
+
+        cc = ContextCollection()
+
+        assert list(cc) == []
+        with cc.added(10):
+            assert list(cc) == [10]
+            with cc.added(20):
+                assert list(cc) == [10, 20]
+            assert list(cc) == [10]
+        assert list(cc) == []
+    """
+    def __init__(self):
+        self._items = PythonOrderedDict()
+        self._index_generator = count(1)
+
+    def __iter__(self):
+        return iter(self._items.values())
+
+    @contextmanager
+    def added(self, value):
+        index = next(self._index_generator)
+        try:
+            self._items[index] = value
+            yield index
+        finally:
+            del self._items[index]
--- a/easypy/decorations.py
+++ b/easypy/decorations.py
@@ -234,3 +234,28 @@ def lazy_decorator(decorator_factory):
     def wrapper(func):
         return LazyDecoratorDescriptor(decorator_factory, func)
     return wrapper
+
+
+class RegistryDecorator(object):
+    """
+    A factory for simple decorators that register functions by name::
+
+        register = RegistryDecorator()
+
+        @register
+        def foo():
+            return 'I am foo'
+
+        @register
+        def bar():
+            return 'I am bar'
+
+        assert register.registry['foo']() == 'I am foo'
+        assert register.registry['bar']() == 'I am bar'
+    """
+    def __init__(self):
+        self.registry = {}
+
+    def __call__(self, fn):
+        self.registry[fn.__name__] = fn
+        return fn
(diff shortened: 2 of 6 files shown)


idanarye:
@koreno, @DoronCohen - I need your opinion.

I'm going to need to add some configuration fields on the fields, to allow customizing some of `TypedStruct` functionalities. Specifically I have two in mind right now:

* I want to add `__hash__` to `TypedStruct`, but there may be fields that are unhashable by default. I need to be able to tell `TypedStruct` to ignore them - or even to specify a custom hash function for them:
	```python
	class Foo(TypedStruct):
		name = str
		data = [str]
		data.hash = None  # disable hashing for this field

	# Or

	class Foo(TypedStruct):
		name = str
		data = [str]
		data.hash = lambda data: hash(tuple(data))  # convert to tuple and then hash
	```
* Some fields are too big to be printed in `__repr__~, so I want to be able to either omit them or create custom print function for them:
	```python
	class Foo(TypedStruct):
		name = str
		data = [str]
		data.repr = None  # disable printing this field in __repr__

	# Or

	class Foo(TypedStruct):
		name = str
		data = [str]
		data.repr = lambda data: '%s items' % len(data)  # just print the number of items
	```

Now, regarding the new syntax - I'm pondering how to approach it:

1. Give them their own context managers:
	```python
	class Foo(TypedStruct):
		name = str
		with HASH(None), REPR(NONE):
			data = [str]
	```
	This will be consistent, but I'm not sure if I want to start giving a context manager for each and every configuration (there may be more in the future)
2. Have a single context manager `CONFIG` with keyword arguments:
	```python
	class Foo(TypedStruct):
		name = str
		with CONFIG(hash=None, repr=None):
			data = [str]
	```
	If I pick this approach - should I merge `DEFAULT` into it as well? I think `DEFAULT` deserve it's own context manager...
3. Don't give them context managers, let users do it manually with `APPLY`:
	```python
	def set_hash_and_repr(field):
		field.hash = None
		field.repr = None

	class Foo(TypedStruct):
		name = str
		with APPLY(set_hash_and_repr):
			data = [str]
	```
	This is also an option, that might be enough for the beginning...

idanarye:
How do you feel about this syntax?

```python
class Foo(TypedStruct):
    a = int; DEFAULT(5)
    b = float; CONVERTIBLE_FROM(int, str)

    # Or, if there are many fields to set:
    c = str
    DEFAULT(20)
    CONVERTIBLE_FROM(int, str)
    VALIDATION(lambda s: 2 <= len(s))
```

idanarye:
> I like the approach but wondering if it should be assigned with `NotImplemented` instead of `None`.

`NotImplemented` feels like more "if someone tries to use this throw an exception" than a "skip this one".

DoronCohen:
I can bare with the semicolon syntax although I don't think linters like the use of it. second option is not optimal, maybe use indentation in some way like:
    class Foo(TypedStruct):
        c = str
            DEFAULT('20')
            ...

you may have to use a tuple for that

idanarye:
If I used a tuple, completion would no longer work. That was also the problem with using `ts.Field`.

PR #78, opened 2018-05-24
Trick completion of `typed_struct.Field`

idanarye:
With this, completion engines should know how to complete the members of `TypedStruct` fields defined with `Field`:

```python
import easypy.typed_struct as ts


class Foo(ts.TypedStruct):
    my_int = ts.Field(int, default=1)


class Bar(ts.TypedStruct):
    my_foo = ts.Field(Foo, default=Foo(my_int=2))


bar = Bar()
bar.my_foo.my_  # with this commit, completion engines can complete this
```

I've tested this with [Jedi](http://jedi.readthedocs.io), which is the Python completion backend used by most editors. PyCharm is using it's own completion engine - @tigrawap, can you check that it works there too?


tigrawap:
IT does not

screen shot 2018-05-24 at 2 02 11 pm

tigrawap:
Not really


screen shot 2018-05-24 at 2 37 16 pm
screen shot 2018-05-24 at 2 36 36 pm

idanarye:
OK, I've installed PyCharm and did some experimenting. It looks like two things are preventing it from performing the completion:

1. ```python
    assert False, 'this function should never actually be called'
    ````
    Apparently PyCharm is smarter than Jedi here, and figures that the function is not going to actually return anything.
2. Apparently PyCharm does not look at the body of functions imported from other modules? The completion here works:
    ```python
    import easypy.typed_struct as ts
    from easypy.typed_struct import Field


    __field_class = Field
    def Field(type, *args, **kwargs):
        if isinstance(type, list):
            return [type[0]()]
        if isinstance(type, dict):
            (k, v), = type.items()
            return {k(): v()}
        return type()
    locals()['Field'] = __field_class



    class Foo(ts.TypedStruct):
        my_int = Field(int, default=1)


    class Bar(ts.TypedStruct):
        my_foo = Field(Foo, default=Foo(my_int=2))


    bar = Bar()
    bar.my_foo.my
    ```

The first problem is easy to deal with - I can just remove the `aasert`. It's not strictly needed, certainly not if it hampers the purpose of this code.

I don't see how the second problem can be solved...

PR #88, opened 2018-06-14
Close #80: Add `EasyMeta` (for now with only one hook)

### LINKED ISSUE #80: The case for EasyMeta
We use a lot of metaclasses:

| Meta Class | What it does |
| -------------- | ---------------- |
| `easypy.concurrency.MultiExceptionMeta` | Generate subtypes with generics syntax |
| `easypy.concurrency.SynchronizedSingleton` | Make all instances of each subclass share the same object, thread safe |
| `easypy.typed_struct.TypedStructMeta` | DSL for declaring typed structs |
| `wepy.aws.instance_types.InstanceTypeMeta` | Not really sure... @koreno / @DoronCohen? |
| `wepy.devops.cluster.setup.ReplaceWithOuter` | @tigrawap? |
| `wepy.devops.jira.JiraProjectMeta` | Register the subclasses |
| `wepy.devops.roles.RoleMeta` | Make subclasses behave like objects, not types |
| `wepy.devops.roles.HostRoleMeta` | Inherits `RoleMeta` and does some trivial setup on top of it |
| `wepy.devops.triage.TriagePluginMeta` | Register the subclasses (unless a `SHOULD_RUN` field is set) |
| `wepy.explore.faults.FaultsMeta` | Decorate methods |
| `qa.mayhem.mayhem_actions.MayhemActionMeta` | Register the subclasses (unless they are abstract) and does some field verification and auto-setting |
| `qa.testlight.system_setup._ABCMetaWithLRUCache` | Cache all instances of each subclass constructed with the same parameters |
| `qa.testlight.utils.ParametrizedBehaviorFixture` | DSL for creating a pytest fixture (not a type!) |

`ParametrizedBehaviorFixture` does not create a type, so it should remain it's own metaclass.

`RoleMeta`/`HostRoleMeta` are data notation - they would benefit from the discussion at https://github.com/weka-io/black-magic/issues/1

For everything else on the list (except `InstanceTypeMeta` and `ReplaceWithOuter`, which I didn't really understand what they do), I want to create a new module - `easypy.meta` - with a metaclass named `EasyMeta` that will contain all these features:
* Automatic class registration
* Subclass verification/modification
* Automatic method decoration
* Easy DSL definitions

The reason for using a single metaclass is that multi-inheritance does not work well with different metaclasses.

----------------

My grepping also found these:
```
$ git grep 'a metaclass could make this go away...'

wepy/devops/jobs/coverage/__init__.py:TestDefinitionRow.table_cls.DATA_CLS = TestDefinitionRow  # a metaclass could make this go away...
wepy/devops/jobs/coverage/__init__.py:JobExecutionRow.table_cls.DATA_CLS = JobExecutionRow  # a metaclass could make this go away...
wepy/devops/jobs/coverage/__init__.py:TestExecution.table_cls.DATA_CLS = TestExecution  # a metaclass could make this go away...
```
We can probably add that functionality to `EasyMeta` too.

idanarye:
Also add first mixin (constructor) that uses `EasyMeta`: `auto_registrar`.

This is a proof of concept for combinable meta mixins - in the future we will add more hooks that'll allow more mixins.

The syntax is a bit different than what I wrote in #80. I'm decorating the hooks with `EasyMeta.Hook`, and making the DSL (the dictionary used in the class declaration) "swallow" them - so that they won't appear in the actual class. This also means we don't need to fear several mixins using the same hook overwriting each other's work. But this also means I'll have to do something about hooks we can only have one of (e.g. hooks that replace instance creation). Still - better than having these hooks decided arbitrarily by the order of baseclass resolution...

### DIFF
--- a/easypy/meta.py
+++ b/easypy/meta.py
@@ -1,0 +1,104 @@
+from abc import ABCMeta
+from functools import wraps
+from collections import OrderedDict
+
+from .decorations import kwargs_resilient
+
+
+class EasyMeta(ABCMeta):
+    """
+    Base class for various meta-magic mixins.
+
+    Use the hooks in :class:`EasyMetaHooks`, decorated with `@EasyMeta.Hook`,
+    to add functionality. Multiple hooks with the same name can be defined,
+    and they will all be invoked sequentially.
+    """
+
+    @classmethod
+    def __prepare__(metacls, name, bases, **kwds):
+        dsl = EasyMetaDslDict()
+        return dsl
+
+    class Hook(object):
+        def __init__(self, dlg):
+            self.dlg = dlg
+
+    def __init__(cls, name, bases, dct, **kwargs):
+        super().__init__(name, bases, dct)
+
+    def __new__(mcs, name, bases, dct, **kwargs):
+        hooks = EasyMetaHooks(class_kwargs=kwargs)
+
+        for base in bases:
+            if isinstance(base, EasyMeta):
+                hooks.extend(base._em_hooks)
+
+        new_type = super().__new__(mcs, name, bases, dct)
+
+        new_type._em_hooks = hooks
+
+        hooks.after_subclass_init(new_type)
+
+        hooks.extend(dct.hooks)
+
+        return new_type
+
+
+class EasyMetaHooks:
+    """
+    Hooks for ``EasyMeta``
+    """
+
+    HOOK_NAMES = []
+
+    def hook(dlg, HOOK_NAMES=HOOK_NAMES):
+        HOOK_NAMES.append(dlg.__name__)
+
+        @wraps(dlg)
+        def hook(self, *args, **kwargs):
+            kwargs_resilience = kwargs_resilient(negligible=self.class_kwargs.keys())
+            kwargs.update((k, v) for k, v in self.class_kwargs.items() if k not in kwargs)
+
+            for hook in self.hooks[dlg.__name__]:
+                kwargs_resilience(hook)(*args, **kwargs)
+
+        return hook
+
+    def __init__(self, class_kwargs={}):
+        self.hooks = {name: [] for name in self.HOOK_NAMES}
+        self.class_kwargs = class_kwargs
+
+    def add(self, hook):
+        self.hooks[hook.__name__].append(hook)
+
+    def extend(self, other):
+        for k, v in other.hooks.items():
+            self.hooks[k].extend(v)
+
+    @hook
+    def after_subclass_init(self, cls):
+        """
+        Invoked after a subclass is being initialized
+
+        >>> class PrintTheName(metaclass=EasyMeta):
+        >>>     @EasyMeta.Hook
+        >>>     def after_subclass_init(cls):
+        >>>         print('Declared', cls.__name__)
+        >>>
+        >>>
+        >>> class Foo(PrintTheName):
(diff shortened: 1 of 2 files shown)


idanarye:
After discussing with @DoronCohen, we decided to remove `auto_registrar` for now, for the following reasons:

1. It's syntax is non-intuitive.
2. It needs far too much customization.
3. It's easy to write on-the-spot in places that use it.

If we ever want to restore it, here is it's code:

```python
def auto_registrar(target, cond=lambda cls: True):
    """
    Create a mixin that automatically registers its subclasses.

    :param target: Where to register the subclasses:

                   * When ``list``, the subclass will be appended to the list.
                   * When ``set``, the subclass will be added to the set.
                   * When ``dict``, the subclass will be added to the dict with the class name as key.
                   * When ``callable``, the callable will be called with the subclass.
    :param cond: Predicate to decide if a subclass will be registered

    >>> l = []
    >>> s = set()
    >>> d = {}
    >>> p = lambda cls: print('Registering', cls.__name__)
    >>>
    >>> RegisterL = auto_registrar(l)
    >>> RegisterS = auto_registrar(s)
    >>> RegisterD = auto_registrar(d)
    >>> RegisterP = auto_registrar(p)
    >>>
    >>> class Foo(RegisterL, RegisterS, RegisterD, RegisterP):
    >>>     pass
    Registering Foo
    >>> l
    [__main__.Foo]
    >>> s
    {__main__.Foo}
    >>> d
    {'Foo': __main__.Foo}

    Subclasses can set ``dont_register=True`` in their declaration header to
    bypass registration just for this class:

    >>> FOOS = []
    >>>
    >>> class Foo(auto_registrar(FOOS), dont_register=True):
    >>>     pass
    >>>
    >>> class Bar(Foo):
    >>>     pass
    >>>
    >>> class Baz(Bar, dont_register=True):
    >>>     pass
    >>>
    >>> class Qux(Baz):
    >>>     pass
    >>>
    >>> FOOS
    [__main__.Bar, __main__.Qux]
    """
    if isinstance(target, list):
        def registration_function(cls):
            target.append(cls)
    elif isinstance(target, set):
        def registration_function(cls):
            target.add(cls)
    elif isinstance(target, dict):
        def registration_function(cls):
            target[cls.__name__] = cls
    else:
        assert callable(target), 'target must be list/set/dict/callable'
        registration_function = target

    dct = EasyMetaDslDict()

    @dct.hooks.add
    def after_subclass_init(cls, dont_register=False):
        if not dont_register and cond(cls):
            registration_function(cls)

    return EasyMeta('<auto_registrar>', (), dct)
```

And it's test:
```python
def test_auto_registrar():
    xs = []
    ys = {}
    zs = set()
    ws = []

    X = auto_registrar(xs)
    Y = auto_registrar(ys, cond=lambda cls: not getattr(cls, 'dont_register', False))
    Z = auto_registrar(zs)
    W = auto_registrar(lambda cls: ws.append(cls.__name__))

    class Foo(X, W):
        pass

    class Bar(Foo, dont_register=True):
        pass

    class Baz(Bar, Y):
        pass

    class Qux(Baz, Z):
        dont_register = True

    assert xs == [Foo, Baz, Qux]
    assert ys == {'Baz': Baz}
    assert zs == {Qux}
    assert ws == ['Foo', 'Baz', 'Qux']
```

PR #89, opened 2018-06-20
BREAKING CHANGE - concurrency: make MultiObject immutable; remove list operators

koreno:
We're doing it!

### DIFF
--- a/easypy/concurrency.py
+++ b/easypy/concurrency.py
@@ -602,7 +602,7 @@ class MultiObject(object, metaclass=MultiObjectMeta):
     CONCESTOR = object
 
     def __init__(self, items=None, log_ctx=None, workers=None, initial_log_interval=None):
-        self._items = list(items) if items else []
+        self._items = tuple(items) if items else ()
         self._workers = workers
         self._initial_log_interval = initial_log_interval
         cstr = self.CONCESTOR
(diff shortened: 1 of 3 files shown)


easypy/concurrency.py
```diff
@@ -574,23 +574,23 @@ def concurrent_map(func, params, workers=None, log_contexts=None, initial_log_in
 class MultiObject(object):
 
     def __init__(self, items=None, log_ctx=None, workers=None, initial_log_interval=None):
-        self._items = list(items) if items else []
+        self._items = tuple(items if items else [])
```
idanarye: `tuple(items) if items else ()`?

easypy/concurrency.py
```diff
@@ -601,7 +601,7 @@ def __init__(self, items=None, log_ctx=None, workers=None, initial_log_interval=
 
     @property
     def L(self):
-        return list(self)
+        return self._items
```
idanarye: I'm actually concerned about this - one would expect `.L` to return a **L**ist. How about keeping `.L` and adding `.T` for retrieving the tuple?

easypy/concurrency.py
```diff
@@ -640,62 +640,19 @@ def do_it(obj, **more_kwargs):
         return self._new(ret)
 
     def __dir__(self):
-        return sorted(set.intersection(*(set(dir(obj)) for obj in self)))
+        return sorted(set.intersection(*(set(dir(obj)) for obj in self))) + super().__dir__()
```
idanarye: Is it legal for `__dir__` to return the same object twice?

easypy/concurrency.py
koreno: I agree, though we'd want to search&replace L->T because normally we don't need it as a mutable list.

easypy/concurrency.py
koreno: I agree, though we'd want to search&replace L->T because normally we don't need it as a mutable list.

PR #91, opened 2018-07-08
Add a `EasyMixin` and `UninheritedDefaults`

idanarye:


idanarye:
https://github.com/weka-io/easypy/blob/easy-meta-mixin-framework-and-UninheritedDefaults/tests/test_meta.py#L81

PR #92, opened 2018-07-09
Add `repr` and `hash` to typed struct fields

idanarye:
* `repr` can configure how the field in represented in textual
  representation of the typed struct.
* `hash` can configure how the field is considered when hashing the
  typed struct.


idanarye:
Motivation: see https://bitbucket.org/wekaio/wekapp/src/01a0222234140027f7ead3424e00e2781af32c51/qa/mayhem/mayhem_actions.py#lines-1620:1637

With this PR, I can easily hide `samples` (without wrapping it in a new class) and remove if from the hashing (it's not really needed there, and we can't hash a dict)

PR #94, opened 2018-07-10
Add `annotate_exceptions`

idanarye:


idanarye:
There are many places where this will be useful, but specifically I want to use it in https://wekaio.atlassian.net/browse/WEKAPP-68225:

```python
def _cloud_rpc(self, method, *params_list, **params_dict):
	    with annotated_exceptions() as annotate:
		    # ...

		    if ":" in parsed.netloc:
			    host, port = parsed.netloc.split(":")
		    else:
			    host, port = parsed.netloc, None
		    annotate(host=host, port=port)

		    # ...

		    annotate(path=path)

		    @retrying(times=3, acceptable=(ConnectionError, OSError), sleep=1, max_sleep=5)  # this is to deal with sporadic disconnections
		    @retrying(times=5, acceptable=(httpclient.HTTPException, HTTPException), sleep=5, max_sleep=10)
		    @retrying(times=2, acceptable=RetrySignal, sleep=0)  # this is in case a Retry was intentionally requested
		    @retrying(times=6, acceptable=exceptions.CloudIdentityTokenAcquireFailure, sleep=5)  # retry if portal is unavailable
		    def perform_cloud_request():
			    # ...

		    return perform_cloud_request()
```

With this, even builtin exceptions will be converted to something we can annotate.

koreno:
If you fix the conflicts I think we should merge this

PR #95, opened 2018-07-17
exceptions/logging: add aborted exception base class

DoronCohen:


easypy/exceptions.py
```diff
@@ -133,6 +133,13 @@ def make(cls, name, template):
         return type(name, (cls,), dict(template=template))
 
 
+class AbortedException(Exception):
```
koreno: This is a logging 'tool'. Even though it is an exception, it does not belong in the exceptions module. The exception module should contain code that helps and improves exceptions in python.

It should be in `easypy.logging`.

koreno: I would also inherit from `BaseException`, to somewhat indicate that this isn't a real exception to be raised on its own, but that it should be inherited from by real exceptions.

PR #97, opened 2018-07-19
concurrency: add a pointer to the exception type the ME is a type of;…

koreno:
… support ME pickling

### DIFF
--- a/easypy/concurrency.py
+++ b/easypy/concurrency.py
@@ -17,7 +17,7 @@ import os
 import signal
 import threading
 import time
-import warnings
+from collections import namedtuple
 from datetime import datetime
 
 from easypy.decorations import parametrizeable_decorator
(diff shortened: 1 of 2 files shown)


idanarye:
I propose to replace the test with something like this:

```python
def test_multiexception_pickling():
    import pickle
    import multiprocessing

    def fail_and_dump(queue):
        try:
            MultiObject(range(5)).call(_ExceptionForPicklingTest.throw)
        except MultiException[ArithmeticError] as exc:
            p = pickle.dumps(exc)
            queue.put_nowait(p)

    queue = multiprocessing.Queue(1)
    process = multiprocessing.Process(target=fail_and_dump, args=(queue,))
    process.start()
    process.join()
    p = queue.get_nowait()

    exc = pickle.loads(p)
    assert isinstance(exc, MultiException[_ExceptionForPicklingTest])
    assert exc.common_type is _ExceptionForPicklingTest
    assert exc.exceptions[0].args == (0,)
    assert exc.exceptions[1:] == [None, None, None, None]
```

By using the `multiprocessing` module we can simulate the multiprocess environment this feature is designed for without depending on binary literals.

PR #99, opened 2018-08-02
decorations: support params for singleton_contextmanager

koreno:


koreno:
`lru_cache` has that problem too... could it be resolved with `Signature.bind`

PR #100, opened 2018-08-08
Add `exception_to_boolean`

idanarye:
I'm not really sure about the name, and I'm also not sure if `resilience` is the right place for this.

The ultimate goal of this is to allow easy transition to `PredicateNotSatisfied`. We have many places in our code that check if something is alive, or if certain conditions are met, and other places that `wait` on these functions. Making the functions that check them raise a `PredicateNotSatisfied` with relevant information would help us debug these timeouts, but because these methods are used all over the place it is too risky to make them throw instead of returning `False`.

By decorating these functions with `@exception_to_boolean` we could provide that information while still allowing to use them as simple predicates. For example, we have `WekaNode.is_alive()`:
```python
    def is_alive(self):
        return self.machine.is_alive() and bool(self.get_proc_info().pid)
```

Which calls `WekaMachine.is_alive()`:
```python
    def is_alive(self, timeout=5):
        with resilience():
            for _ in self.cmd.echo.popen().iter_lines(line_timeout=5):
                return True
        return False
```

These could be written as:

```python
# The one in WekaMachine

    @exception_to_boolean(acceptable=PredicateNotSatisfied)
    def is_alive(self, timeout=5):
        with resilience():
            for _ in self.cmd.echo.popen().iter_lines(line_timeout=5):
                return True
        raise PredicateNotSatisfied('{machine} did not respone', machine=self)

# The one in WekaNode

    @exception_to_boolean(acceptable=PredicateNotSatisfied)
    def is_alive(self):
        self.machine.is_alive.func()
        proc_info = self.get_proc_info()
        if not proc_info.pid:
            raise PredicateNotSatisfied('Process of {node} is dead. Proc info = {proc_info}', node=self, proc_info=proc_info)
        return True
```

(maybe we'll want to make `func()` swallow the returned value of the underlying function and **always** return `True` if it didn't throw)

With this, we can easily call `is_alive()` nodes and hosts and get a boolean result, but when `wait`ing on them we could call `is_alive.func()` to get a more detailed exception in case of timeout.

### DIFF
--- a/easypy/resilience.py
+++ b/easypy/resilience.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 
-from functools import partial, wraps
+from functools import partial, wraps, update_wrapper
 from contextlib import contextmanager
 from easypy.timing import Timer
 from easypy.units import Duration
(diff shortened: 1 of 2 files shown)


koreno:
I suggest to put this in `wepy.devops.utils`, as it isn't really a standalone utility in easypy except for this transition

koreno:
(but I like the idea!)

koreno:
you might be able to justify a general purpose `converts_exceptions` decorator, which accepts an exception type and a lambda that tells it how to convert these exceptions into return values.

general comment - I'd like to put more energy into refining the existing easypy stuff (and remove clutter and unusables), rather than add more and more features, for the time being.

PR #101, opened 2018-08-21
Signals handlers keep methods as weak references

idanarye:
This makes signal handlers use weak references to methods so that they won't be keeping the objects alive.

Questions:
1. Do we want an opt-out argument?
2. Do we want to, instead of making `Signal.register()` convert methods to `WeakMethod`, have the conversion be done in `register_object` so that directly registered methods won't be weakly referenced?
    * On the one hand, these are methods too and we don't want the globally scoped signal to keep their objects alive.
    * On the other hand, one would usually register these methods with the `registered` context manager, so unregistration will be done automatically and weakreffing could be potentially harmful.

### DIFF
--- a/easypy/signals.py
+++ b/easypy/signals.py
@@ -6,9 +6,10 @@ from itertools import chain, count
 import logging
 from enum import Enum
 from contextlib import contextmanager, ExitStack
+import weakref
 
 from easypy.concurrency import Futures, MultiObject
-from easypy.decorations import parametrizeable_decorator, kwargs_resilient
+from easypy.decorations import parametrizeable_decorator, kwargs_resilient, WeakMethodDead
 from easypy.exceptions import TException
 from easypy.contexts import is_contextmanager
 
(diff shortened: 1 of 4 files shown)


easypy/signals.py
```diff
@@ -188,18 +196,29 @@ def __call__(self, **kwargs):
...
                 futures = None
 
+            handlers_to_remvoe = []
+
+            def run_handler(handler):
+                try:
+                    with _logger.context("#%03d" % handler.idx):
+                        handler(**kwargs)
+                except WeakMethodDead:
+                    handlers_to_remvoe.append(handler)
```
DoronCohen: typo

easypy/signals.py
```diff
@@ -312,16 +354,32 @@ def register_object(obj, **kwargs):
...
 def unregister_object(obj):
-    for method_name in get_signals_for_type(type(obj)):
-        method = getattr(obj, method_name)
-        if not callable(method):
-            continue
-        method_name, *_ = method_name.partition("__")
-        unregister_signal(method_name, method)
+    for signal_name in {method_name.split('__', 1)[0] for method_name in get_signals_for_type(type(obj))}:
+        for handlers in Signal.ALL[signal_name].handlers.values():
+            handlers[:] = (handler for handler in handlers if handler.bound_object is not obj)
```
idanarye: I've changed this because I think the current object unregistration may be too slow - see https://wekaio.atlassian.net/browse/WEKAPP-73925

PR #102, opened 2018-08-25
Putting functions into the appropriate modules

koreno:


koreno:
(This is where the code is consolidated; indeed some conflicts need to be resolved)

PR #103, opened 2018-08-28
`singleton_contextmanager` remove from cache on `__exit__`

idanarye:
This prevents the `singleton_contextmanager` from keeping it's arguments
alive (most importantly - `self` for methods) even after the
contextmanager has finished.

### DIFF
--- a/easypy/decorations.py
+++ b/easypy/decorations.py
@@ -73,34 +73,34 @@ def parametrizeable_decorator(deco):
 def singleton_contextmanager(func):
     from .caching import locking_cache
 
+    class CtxManager():
+        def __init__(self, *args, **kwargs):
+            self.args = args
+            self.kwargs = kwargs
+            self.count = 0
+            self.func_cm = contextmanager(func)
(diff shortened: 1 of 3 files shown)


easypy/decorations.py
```diff
@@ -73,34 +73,34 @@ def inner(func=None, **kwargs):
 def singleton_contextmanager(func):
     from .caching import locking_cache
 
+    class CtxManager():
```
idanarye: I moved it outside the `inner` class - not sure why was it even there in the first place.

Should it be moved to the global scope entirely?

koreno: perhaps... "ReentrantContextManager"?...
Does it have any justification on its own?

idanarye: The name of the class should follow the name of the function - so `SingletonContextManager`, and if we change the name later we'll change the name of the class as well.

As for justifications - I don't think one needs justifications to have a class in the global scope. This should be the default approach, and you need a special reason to declare a class inside a function.

PR #106, opened 2018-09-02
in our automation (bamboo/reggie) PersistentCaching is disabled due to some long forgotten bug,

koreno:
and we need this to get some caching going at least within the process ('_get_instance_types_meta', for example)

### DIFF
--- a/easypy/caching.py
+++ b/easypy/caching.py
@@ -141,12 +141,17 @@ class PersistentCache(object):
             ret = func(*args, **kwargs)
             self.set(key, ret)
             return ret
+
+        inner = timecache(expiration=self.expiration)(inner)
+        self._timecache_clear = inner.cache_clear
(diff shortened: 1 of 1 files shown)


easypy/caching.py
```diff
@@ -141,12 +141,17 @@ def inner(*args, **kwargs):
             ret = func(*args, **kwargs)
             self.set(key, ret)
             return ret
+
+        inner = timecache(expiration=self.expiration)(inner)
+        self._timecache_clear = inner.cache_clear
```
idanarye: Maybe I'm not following this correctly, but wouldn't this override `self._timecache_clear` to be the `cache_clear` of the _last_ returned cache? If the `PersistentCache` is called multiple times, `clear()` will only clear the cache of the last call?

PR #110, opened 2018-09-20
Prevent `timecache` from leaking when used with methods

idanarye:


idanarye:
Companion wekapp ticket: https://bitbucket.org/wekaio/wekapp/pull-requests/5647

PR #116, opened 2018-10-09
logging: support for suppressing and soloing logging to console per thread

koreno:
wanted this to suppress logging from bg thread while AppMenu is showing its UI

### DIFF
--- a/easypy/logging.py
+++ b/easypy/logging.py
@@ -10,7 +10,8 @@ import time
 import traceback
 from contextlib import ExitStack
 from functools import wraps, partial
-from itertools import cycle, chain, repeat
+from itertools import cycle, chain, repeat, count
+from collections import OrderedDict
 
 from easypy.colors import colorize_by_patterns as C, uncolorize
 from easypy.humanize import compact as _compact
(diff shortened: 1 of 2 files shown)


easypy/logging.py
```diff
@@ -126,6 +126,47 @@ def get_console_handler():
...
+    @classmethod
+    @contextmanager
+    def solo(cls):
+        try:
+            with cls.LOCK:
+                cls.SELECTED.append(threading.current_thread())
+            yield
+        finally:
+            with cls.LOCK:
+                cls.SELECTED.pop(-1)
```
idanarye: I'm not that familiar with the logging configuration, but is it safe to do something like this to a list shared between threads?

easypy/logging.py
```diff
@@ -126,6 +126,47 @@ def get_console_handler():
...
+
+    def filter(self, record):
+        selected = self.SELECTED[-1]
+        if selected:
```
idanarye: So this is why `SELECTED` is initialized to `[None]`? Wouldn't it be simpler to set it to `[]` and do this:
```python
try:
    selected = self.SELECTED[-1]
except IndexError:
    return not self.CONTEXT.silenced
else:
    return selected == threading.current_thread()
```

easypy/logging.py
```diff
@@ -126,6 +127,52 @@ def get_console_handler():
...
+        try:
+            with cls.LOCK:
+                idx = next(cls.IDX_GEN)
+                cls.SELECTED[idx] = threading.current_thread()
+            yield
+        finally:
+            cls.SELECTED.pop(idx)
+
+    def filter(self, record):
+        selected = self.SELECTED[self.SELECTED._OrderedDict__root.prev.key]  # hacky way for quickly getting to the last element
```
idanarye: This will not work in newer versions of Python (3.5+ I think). `next(reversed(self.SELECTED))` will get you the last key and is also _O(1)_ even in 3.4.

koreno: you don't say...

PR #136, opened 2018-12-04
signals: bug fix - make sure to invoke sync/async handlers in their r…

koreno:
…espective priority batch

### DIFF
--- a/easypy/signals.py
+++ b/easypy/signals.py
@@ -163,6 +163,10 @@ class Signal:
     def iter_handlers(self):
         return chain(*(self.handlers[k] for k in PRIORITIES))
 
+    def iter_handlers_by_priority(self):
+        for k in PRIORITIES:
+            yield separate(self.handlers[k], lambda h: h.async)
+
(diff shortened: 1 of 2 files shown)


easypy/signals.py
```diff
@@ -301,24 +314,15 @@ def run_handler(handler):
...
-                else:
-                    handlers.append(handler)
-            if async_handlers:
-                handlers_stack.enter_context(async_handlers.call(run_handler))
-            for handler in handlers:
-                res = run_handler(handler)
-                if res:
-                    handlers_stack.enter_context(res)
+            for async_handlers, synced_handlers in self.iter_handlers_by_priority():
+                async_handlers = MultiObject(filter(should_run, async_handlers))
```
idanarye: Why not just put the `filter` in `iter_handlers_by_priority`?
```python
def iter_handlers_by_priority(self):
    for k in PRIORITIES:
        yield separate(filter(should_run, self.handlers[k]), lambda h: h.async)
```

PR #139, opened 2018-12-09
Fix #138: reorganize multithreading init code

### LINKED ISSUE #138: Reorganize multithreading init code
`easypy.__init__` defines some methods and functions that are needed for some multithreading modules like `easypy.threadtree`, `easypy.concurrency` or `easypy.gevent`. We want to move these to their own module (their rightful place is `easypy.threadtree`, but putting them there would create import cycles) and leave `__init__.py` empty (later we will add a documentation of the whole package to there)

idanarye:


idanarye:
@koreno - can I merge this?

PR #143, opened 2019-01-02
collections: added 'takesome'

koreno:
This came up a bunch of times...

### DIFF
--- a/easypy/collections.py
+++ b/easypy/collections.py
@@ -852,6 +852,45 @@ def as_list(generator, sort_by=None):
     return inner
 
 
+def takesome(generator, max=None, min=0):
+    """
+    Take between ``min`` and ``max`` items from the generator.
+    Throws a ValueError if the generator didn't yield ``min`` items.
+
+    >>> list(takesome("abcdef", 3))
+    ['a', 'b', 'c']
+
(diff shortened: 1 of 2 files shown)


easypy/collections.py
```diff
@@ -847,6 +847,39 @@ def inner(*args, **kwargs):
...
+    """
+
+    items = [p for p, _ in zip(generator, range(min))]
+
+    if len(items) < min:
+        raise ValueError("Not enough items in sequence (wanted {}, got {})".format(min, len(items)))
+
+    yield from items
+
+    if not max:
```
idanarye: I think `max=0` should yield 0 items, not (potentially) infinite items. This can bite us if we ever do something like:
```python
yield from buffer  # buffer was filled with stuff before
yield from takesome(data_source, chunk_size - buffer)
```
This snippet will work as long as `buffer` does not contain exactly the number of items we want to send in each chunk.

DoronCohen: what about a couple of uts?

koreno: The docstrings are doctests...

koreno:
all fixed, also added the doctests to the travis ci build

Issue #142, opened 2018-12-30
improve error message from easypy's collections


koreno:
collections should support a .name attribute, that can be used in `repr`s. Calculated collections (such as `FilterCollection`) should derive a name from their parent collection.
This `repr` should be used in `ObjectNotFound`, which itself should be upgraded into a `TException`.

DoronCohen:
how about `ListCollection[Machine]` like in `MultiObject` on `__repr__`?

PR #148, opened 2019-01-25
collections: raise informative exceptions when object-not-found (#142)

koreno:
- support naming collections
- improve repr
- support ipython pretty repr

### DIFF
--- a/easypy/collections.py
+++ b/easypy/collections.py
@@ -8,10 +8,12 @@ import collections
 from numbers import Integral
 from itertools import chain, islice
 from functools import partial
+import inspect
 import random
 from .predicates import make_predicate
 from .tokens import UNIQUE
 from .decorations import parametrizeable_decorator
+from .exceptions import TException
 
 import sys
 if sys.version_info[:2] >= (3, 5):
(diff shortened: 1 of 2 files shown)


easypy/collections.py
```diff
@@ -24,8 +25,38 @@
     from collections import OrderedDict as PythonOrderedDict
 
 
-class ObjectNotFound(LookupError):
-    pass
+def _format_filter_string(preds, filters):
+    returned = []
+    if preds:
+        returned.append("predicates=[%s]" % ", ".join(map(str, preds)))
```
idanarye: This separation seems a bit superficial. 


def x_is_even(obj):
    return obj.x % 2 == 0

print(_format_filter_string([x_is_even], dict(y=5)))


This will print `predicates=[<function x_is_even at 0x7f99923466a8>] filters=[y=5]`. But wouldn't it be better if it'd print: `[x_is_even, y=5]`?

easypy/collections.py
```diff
@@ -532,6 +591,16 @@ def _new(self, items):
...
         return self.base.__class__(items)
 
+    def __repr__(self):
+        if self.name and self.base.name:
+            return "<{0.name}, based on {0.base.name} ({0.__class__.__name__})>".format(self)
+        elif self.name:
+            return "<{0.name}, based on a {0.base.__class__.__name__} ({0.__class__.__name__})>".format(self)
+        elif not self.name:
+            return "<A {0.__class__.__name__} based on {0.base.name}>".format(self)
+        else:
```
yuvalweka: else clause is unreachable.

PR #156, opened 2019-03-13
Deleted randutils module and moved its used functions to misc module

YuvalEvron:


idanarye:
Why are we moving it to `misc` again? I thought #154 was just about removing the `randutils` functions that we don't need.

A module named `misc` is always a compromise for putting things that don't have a natural home. Removing modules with clearly defined name and purpose and shoving their functionality into `misc` should not be a goal...

koreno:
Yes, I agree, I think I'm to blame on this.
Let's then rename this module to `random`, so it conforms to the naming pattern of other modules in easypy (everything is "utils").
`clamp` can be replaced by the `easypy.misc.clamp`, and `interpolate` can be discarded.

PR #157, opened 2019-03-17
Renamed randutils module to random and removed unused XRandom class

YuvalEvron:


easypy/humanize.py
```diff
@@ -16,6 +16,7 @@
 from easypy.bunch import Bunch, bunchify
 from easypy.colors import colorize
 from easypy.collections import ilistify
+from easypy.humanize import clamp
```
idanarye: 1. This already is `easypy.humanize`.
2. Didn't you just delete `clamp` in this very PR? From **both** `easypy.humanize` and `easypy.random`?

easypy/misc.py
```diff
@@ -40,8 +40,10 @@ def stack_level_to_get_out_of_file():
     return stack_levels
 
 
-def clamp(val, *, at_least=None, at_most=None):
-    "Clamp a value so it doesn't exceed specified limits"
+def clamp(val, at_least=None, at_most=None):
```
koreno: no - as discussed, `clamp` should require all 3 params (though they can be passed `None`; which should be documented).

DoronCohen:
feels like it should have been splitted to 2 commits

PR #158, opened 2019-03-17
concurrency: documentation

koreno:


koreno:
Please be spare with doc comments, I want to get the "foot in the door", as long as I didn't introduce bugs...

easypy/concurrency.py
```diff
@@ -1,9 +1,65 @@
...
+This module helps you run things concurrently.
+Most useful are the ``concurrent`` context manager and the ``MultiObject`` class.
+The features in this module are integrated with the ``logging`` module, to provide
+thread-context to log messages. It also has support for integration ``gevent``.
+
+``MultiObject``
+
+Here's how you would, for example, concurrently send a message to a bunch of servers::
+
+    responses = MultiObject(servers).send('Hello')
```
DoronCohen: not accurate. here `responses` is of `MultiObject` type...

koreno: I think that's too pedantic for this area of this documentation. better to keep it simple and broad.

PR #178, opened 2019-07-22
sync: Don't release unacquired lock

YuvalEvron:


DoronCohen:
@YuvalEvron this looks broken in the tests. did you look at them?

idanarye:
Instead of checking if `self.cond.locked()`, we should remove the root bug - the fact the the `acquire` happens inside the `try`. If the `acquire` would happen before the `try` we wouldn't reach the `finally` if we fail to get the lock.

PR #179, opened 2019-07-25
Add semver module

DoronCohen:
We need this module available in other projects. I lost track to where should we push this. `master` / one of the `wekapp` branches. I will also need this in a released package and I saw some effort to release a `0.4.0` what is the status of that?

### DIFF
--- a/easypy/semver.py
+++ b/easypy/semver.py
@@ -22,6 +22,19 @@ class SemVer(namedtuple("SemVer", "major minor patch build tag")):
 
     @classmethod
     def loads(cls, string, *, separator='.', tag_separator='-', raise_on_failure=True):
+        """
+        Load a string into a SemVer object.
+
+        :param string: The string representing the semantic version. Must adhere to semver format.
+        :type string: str
+        :param separator: Use a different version part separator.
+        :type separator: str, optional
+        :param tag_separator: Use a different tag separator.
+        :type tag_separator: str, optional
+        :param raise_on_failure: Whether to raise on failure or just return None
+        :type raise_on_failure: bool, optional
+        """
+
         string, _, tag = string.partition(tag_separator)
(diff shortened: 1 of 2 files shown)


easypy/semver.py
```diff
@@ -0,0 +1,187 @@
...
+            major=int(major),
+            minor=int(minor),
+            patch=int(patch) if patch is not None else None,
+            build=int(build) if build is not None else None,
+            tag="" if tag is None else str(tag),
+        )
+
+    @property
+    def __dict__(self):
+        return super().__dict__
```
idanarye: Why is this necessary?

easypy/semver.py
```diff
@@ -0,0 +1,187 @@
...
+    def _to_tuple(self):
+        return (self.major, self.minor, self.patch or 0, self.build or 0)
+
+    def __eq__(self, other):
+        assert isinstance(other, self.__class__)
+        return self._to_tuple() == other._to_tuple() and self.tag == other.tag
+
+    def __lt__(self, other):
+        assert isinstance(other, self.__class__)
+        return self._to_tuple() < other._to_tuple() or \
```
YuvalEvron: You should add parenthesis to make this statement clear.

easypy/semver.py
```diff
@@ -0,0 +1,187 @@
...
+            self._to_tuple() == other._to_tuple() and self.tag < other.tag
+
+    def __gt__(self, other):
+        assert isinstance(other, self.__class__)
+        return self._to_tuple() > other._to_tuple() or \
```
YuvalEvron: Same

easypy/semver.py
koreno: seems like it doesn't work anyway. I'll remove this

PR #184, opened 2019-08-01
resilience: Add 'unacceptable' arg to 'retrying'

### LINKED ISSUE #168: Add `unacceptable` argument to `retrying` decorator in resilience
`unacceptable` argument can be overridden in `retry` function but not in `retrying`.

YuvalEvron:
Fixes #168

### DIFF
--- a/easypy/resilience.py
+++ b/easypy/resilience.py
@@ -155,7 +155,7 @@ def retry(times, func, args=[], kwargs={}, acceptable=Exception, sleep=1,
             time.sleep(sleep_for)
 
 
-def retrying(times, acceptable=Exception, sleep=1, max_sleep=False, log_level=logging.DEBUG, pred=None):
+def retrying(times, acceptable=Exception, sleep=1, max_sleep=False, log_level=logging.DEBUG, pred=None, unacceptable=()):
     """Try running the decorated function, retrying if an acceptable exception caught.
 
     times - limit number of attempts before errors are propagated instead of suppressed.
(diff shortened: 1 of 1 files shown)


easypy/resilience.py
```diff
@@ -177,11 +178,16 @@ def retrying(times, acceptable=Exception, sleep=1, max_sleep=False, log_level=lo
...
-                pred=pred)
+            custom_args = dict(
+                sleep=sleep,
+                max_sleep=max_sleep,
+                acceptable=acceptable,
+                log_level=log_level,
+                pred=pred
+            )
+            if unacceptable is not AUTO:
+                custom_args.update(unacceptable=unacceptable)  # We want to use 'retry' default value if a value was not specified
```
idanarye: Let's be consistent here. Either have `unacceptable=AUTO` in both `retry` and `retrying` or have `unacceptable=()` in both of them.

easypy/resilience.py
```diff
@@ -177,11 +178,16 @@ def retrying(times, acceptable=Exception, sleep=1, max_sleep=False, log_level=lo
...
-                times, func, args, kwargs,
-                sleep=sleep, max_sleep=max_sleep,
-                acceptable=acceptable, log_level=log_level,
-                pred=pred)
+            custom_args = dict(
+                sleep=sleep,
+                max_sleep=max_sleep,
+                acceptable=acceptable,
+                log_level=log_level,
+                pred=pred
```
idanarye: Nitpick - use either:
```python
    pred=pred,
)
```

Or:
```python
    pred=pred)
```

easypy/resilience.py
YuvalEvron: I disagree, the reason I chose `AUTO` is to prevent a case in which we have to make sure the default value is the same in both `retry` and `retrying`. Since one is using the other, I think setting two default can only cause problems in the future (in case we decide to change the default value or its format).

idanarye: Then use `AUTO` in both.

YuvalEvron: For what purpose? The reason i use `AUTO` in `retrying` is to make `retry`'s default value transparent to it.

YuvalEvron: @DoronCohen - WDYT?

koreno: `retry` and `retrying` should mirror each other's parameters. To the use they are two ways of achieving the same behavior - one by decorating a function (so retrying is a permanent part of its behavior), and the other is for ad-hoc situations, where the use-case requires a some function to be retried. Regardless, their parameter signature should be consistent, and if one uses `AUTO` to specify which exceptions are not allowed, then so should the other.
The fact that one function _uses_ the other should not affect the parameters.

PR #185, opened 2019-08-01
Add the `mirror_defaults` decorator

idanarye:
This can solve the problem of [using AUTO vs replicating arguments](https://github.com/weka-io/easypy/pull/184#discussion_r309612287). We could have:

```python
def retry(times, func, args=[], kwargs={}, acceptable=Exception, sleep=1,
          max_sleep=False, log_level=logging.DEBUG, pred=None, unacceptable=()):
    ...


@mirror_defaults(retry)
retrying(times, acceptable=AUTO, sleep=AUTO, max_sleep=AUTO, log_level=AUTO, pred=AUTO, unacceptable=AUTO):
    ...
```


idanarye:
c.c. @koreno - I can't add you as a reviewer.

koreno:
1. `AUTO` might be a desireable override. better make it `mirror_defaults(foo, token=MIRROR)` (allow using an alternate token, default to `MIRROR`)
2. can this be applied in many other places in easypy? wait/iter_wait? resilient/resilience? if it works then we can consider incorporating this. Otherwise I'm feeling this might be too fringe and magic... (I haven't thought it thoroughly)

koreno:
> c.c. @koreno - I can't add you as a reviewer.

@DoronCohen will I be re-admitted to this repo?...

idanarye:
> * `AUTO` might be a desireable override. better make it `mirror_defaults(foo, token=MIRROR)` (allow using an alternate token, default to `MIRROR`)

How do you feel about `mirror_defaults` being the default "token" (since it's already imported)?

> * can this be applied in many other places in easypy? wait/iter_wait? resilient/resilience? if it works then we can consider incorporating this. Otherwise I'm feeling this might be too fringe and magic... (I haven't thought it thoroughly)

I see no reason why it shouldn't work everywhere - though we may need to have special treatment for generators to avoid messing up reflection...

DoronCohen:
I think that we should allow the users to use these features but using them in the repo might put off other developers from contributing.

koreno:
> I think that we should allow the users to use these features but using them in the repo might put off other developers from contributing.

I think that if we don't use an easypy feature _within_ easypy, then _it shouldn't be in easypy_...
Either we sanction it and demonstrate its greatness in our code, or we deem it improper coding for some reason and therefor reject it.



> > * `AUTO` might be a desireable override. better make it `mirror_defaults(foo, token=MIRROR)` (allow using an alternate token, default to `MIRROR`)
> 
> How do you feel about `mirror_defaults` being the default "token" (since it's already imported)?
> 

mm.... that feels odd and unreadable. `mirror_defaults.MIRROR`?

> > * can this be applied in many other places in easypy? wait/iter_wait? resilient/resilience? if it works then we can consider incorporating this. Otherwise I'm feeling this might be too fringe and magic... (I haven't thought it thoroughly)
> 
> I see no reason why it shouldn't work everywhere - though we may need to have special treatment for generators to avoid messing up reflection...

then I think we should implement this treatment...

idanarye:
> mm.... that feels odd and unreadable. `mirror_defaults.MIRROR`?

Too long. Let's just use `MIRROR` - one could always import it together with `mirror_defaults`

YuvalEvron:
> > mm.... that feels odd and unreadable. `mirror_defaults.MIRROR`?
> 
> Too long. Let's just use `MIRROR` - one could always import it together with `mirror_defaults`

I think that the combination of the two can be confusing. I vote for the current implementation.

idanarye:
Can we decide about the token before I start applying this wherever possible and using that token everywhere?

YuvalEvron:
> Can we decide about the token before I start applying this wherever possible and using that token everywhere?

We can also pass the names of parameter we want to mirrors as `*args* to the decorator.

idanarye:
Another option is to create a decorator that verifies the defaults are the same:

```python
def foo(a=1, b=2, c=3):
    ...


@ensure_same_defaults(foo)
def bar(a=1, b=2, c=4):
    ...
```

This will raise an exception because `bar`'s default for `c` is 4 instead of 3. It's less automatic than `mirror_defaults` but it has the advantage of IPython and IDEs showing the correct signature.

koreno:
So do we still want this? or did we lose interest?

YuvalEvron:
> So do we still want this? or did we lose interest?

Looks like we can't agree on the implementation.

idanarye:
> > So do we still want this? or did we lose interest?
> 
> Looks like we can't agree on the implementation.

Not on the `mirror_defaults` implementation, but I think we can agree on the `ensure_same_defaults` implementation?

YuvalEvron:
> > > So do we still want this? or did we lose interest?
> > 
> > 
> > Looks like we can't agree on the implementation.
> 
> Not on the `mirror_defaults` implementation, but I think we can agree on the `ensure_same_defaults` implementation?

:+1:

PR #187, opened 2019-08-06
signals: allow hiding 'on_*' methods from register_object

koreno:
sometimes we want to `register_object` an existing object that inherits or implements (legacy) `on_` methods that we don't want registered as signal handlers


easypy/signals.py
```diff
@@ -399,6 +422,9 @@ def register_object(obj, **kwargs):
         method = getattr(obj, method_name)
         assert callable(method), "'%s' (%r) is not callable" % (method, obj)
 
+        if getattr(method, "_signals_dont_register", False):
```
idanarye: This worries me a bit, because `Foo.on_bar` is not the same as `Foo().on_bar`. While the regular `__get__` and method type pass this, I'm not sure if it is wise to count on it...

idanarye:
If we inherit these methods, how is this going to solve the problem?

easypy/signals.py
koreno: > If we inherit these methods, how is this going to solve the problem?
I would apply it at the base, or override the methods.

but maybe a better way would be to pass a black-list to `register_object`?
```register_object(obj, ignore=["on_finished", "on_started"])```

Comment (idanarye) [reply easypy/signals.py]: I'm worried about that too - `register_object` is often used generically, without knowing which object we are registering...

Why not just override these methods?
```python
class Foo:
    def on_bar(self):
        ...


class Baz(Foo)(:
    on_bar = None

    def on_qux(self):
        ...


register_object(Baz())  # will not register `on_bar` because it's not a function.
```

koreno:
decided this feature adds complexities that are better resolved on the user side, by separating the objects or renaming existing methods.

PR #188, opened 2019-08-06
collections: workaround python bug importing 'test.support'

koreno:
pretty esoteric, i would remove the sibling methods if we could, its too costly...

### DIFF
--- a/easypy/collections.py
+++ b/easypy/collections.py
@@ -16,14 +16,20 @@ from .decorations import parametrizeable_decorator
 from .exceptions import TException
 
 import sys
+
+from collections import OrderedDict as PythonOrderedDict
+SUPPORT_GET_SIBLING = True
+
 if sys.version_info[:2] >= (3, 5):
     # In order to support 'get_prev' and 'get_next', we need access to OrderedDict's internal .__map,
     # which we don't have in the C-implementation of the class in Python3.5
     # This hack allows us to get to the pythonic implemenation of OrderedDict
-    from test.support import import_fresh_module
-    PythonOrderedDict = import_fresh_module('collections', blocked=['_collections']).OrderedDict
-else:
-    from collections import OrderedDict as PythonOrderedDict
+    try:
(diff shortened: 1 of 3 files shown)


easypy/collections.py
```diff
@@ -553,27 +559,50 @@ def index(self, obj):
...
+        def get_next(self, obj):
+            from warning import warn
+            warn("Couldn't import a pure-python OrderedDict so this will be a O(n) operation")
+
+            from itertools import cycle
+            l1, l2 = cycle(self._objects.values()), cycle(self._objects.values())
+            next(l2)
+            for a, b in zip(l1, l2):
+                if obj == a:
+                    return b
```
YuvalEvron: why not `if obj==a: return next(l1)`  instead of using two iterators?

easypy/collections.py
```diff
@@ -553,27 +559,50 @@ def index(self, obj):
...
+        def get_prev(self, obj):
+            from warning import warn
+            warn("Couldn't import a pure-python OrderedDict so this will be a O(n) operation")
+
+            from itertools import cycle
+            l1, l2 = cycle(self._objects.values()), cycle(self._objects.values())
+            next(l2)
+            for a, b in zip(l1, l2):
+                if obj == b:
+                    return a
```
YuvalEvron: Same as in `get_next`

easypy/collections.py
koreno: The internal implementation is not part of this PR... especially when I'd prefer to remove this code completely.

To your question - It would work here, but not in the `get_prev`, where we need an object we've already passed in the iteration.

easypy/collections.py
koreno: See above

easypy/collections.py
YuvalEvron: Regarding `get_prev`, I'd rather keep the previous item each time instead of iterating twice, but if we are planning to remove this section then it doesn't matter.

PR #189, opened 2019-08-06
Misc updates 0806

koreno:


easypy/colors.py
```diff
@@ -334,7 +336,7 @@ def colorize(text):
 
     def _subfunc(match_obj):
         colorizer = Colorizer.from_markup(match_obj.group(1))
-        return colorizer(match_obj.group(2) or match_obj.group(3))
+        return colorizer(next(filter(None, match_obj.groups()[1:])))
```
YuvalEvron: Is it possible to get a `StopIteration` exception here (less than 2 groups)?

koreno: no, the regex has 4 groups (some may be none).
and the regex wouldn't match if at least one of them isn't populated.

koreno:
@idanarye ?

idanarye:
Still think the change to `DataSize.__str__` is wrong.

koreno:
> Still think the change to `DataSize.__str__` is wrong.

I've compensated with a new `.render` and `.__format__` implementation

PR #191, opened 2019-08-06
Cli tools

koreno:
I never knew about this possibility with `setup.py`! it's pretty useful!

### DIFF
--- a/easypy/ziplog.py
+++ b/easypy/ziplog.py
@@ -155,7 +155,7 @@ if __name__ == "__main__":
     parser = argparse.ArgumentParser(description='ZipLog - merge logs by timestamps')
     parser.add_argument(
         'logs', metavar='N', type=str, nargs='+',
-        help='log files')
+        help='Log files; Use "-" for STDIN')
     parser.add_argument(
         '-i', '--interval', dest='interval', default=None,
         help="Show interval by seconds (s), or milliseconds (ms)")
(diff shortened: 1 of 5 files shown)


easypy/ziplog.py
```diff
@@ -93,11 +144,33 @@ def iter_zipped_logs(*log_streams):
...
+    import argparse
+    parser = argparse.ArgumentParser(description='ZipLog - merge logs by timestamps')
+    parser.add_argument(
+        'logs', metavar='N', type=str, nargs='+',
+        help='Log files; Use "-" for STDIN')
+    parser.add_argument(
+        '-i', '--interval', dest='interval', default=None,
+        help="Show interval by seconds (s), or milliseconds (ms)")
+    parser.add_argument(
+        '-p', '--prefix', dest='prefix', default="> ",
```
idanarye: I don't like that the CLI tool has different defaults than the command (`> ` vs `DARK_GRAY@{> }@`)

koreno: but then are you assuming that the cli would be piping to `ezcolorize`?...
I could add a `--colorize` flag, but that would still mean override the default.

regardless, the default should not use coloring, but then you guys may want to update the ziplog calls in teka.

idanarye: I'd rather change a few calls in `teka` than get stuck with weird arguments...

Also, the worst that could happen from neglecting to change `teka` is not having the marker colored.

PR #197, opened 2019-08-08
concurrency: Add parent's name to new thread's name

### LINKED ISSUE #174: Based unnamed thread names on named threads
There are three "styles" of thread names:
* Unnamed threads - get the name Python assigns to them ("Thread-1234")
* Arbitrarily named threads - created with `concurrent()`, get names like "anon-7F4A1EE18A20".
    * For some reason `concurrent` creates them but `concurrent_map` - which `MultiObject` uses - does not. @koreno - was there a particular reason for that? Why did you rename these threads in the first place?
* Named threads - which we purposefully give names to ("wd-my-watchdog-thread")

Named threads are usually used for macro scale processes (like watchdogs or mayhem actions). The problem is that when they need to "split up" they get arbitrary names making it harder to associate log lines to them.

I think you should use the following rule:

1. If a thread is explicitly named - use that name.
2. Otherwise, go up the thread tree to find a named thread, and append some random/incrementing subidentifier to it ("wd-my-watchdog:AABBCC")
    * Instead of scaling the tree, we can just put the named thread's name in the thread context.
3. If there is no named thread ancestor - use a generic name.

This will improve the ability to read the logs.

c.c. @DoronCohen @YuvalEvron

YuvalEvron:
Fixes #174

### DIFF
--- a/easypy/concurrency.py
+++ b/easypy/concurrency.py
@@ -933,12 +933,19 @@ class concurrent(object):
         self.kwargs = kwargs
         self.throw = kwargs.pop('throw', True)
         self.daemon = kwargs.pop('daemon', True)
-        self.threadname = kwargs.pop('threadname', 'anon-%X' % id(self))
         self.stopper = kwargs.pop('stopper', threading.Event())
         self.sleep = kwargs.pop('sleep', 1)
         self.loop = kwargs.pop('loop', False)
         self.timer = None
         self.console_logging = kwargs.pop('console_logging', True)
+        self.threadname = kwargs.pop('threadname', None)
+        if not self.threadname:
+            current_thread_name = threading.current_thread().name
+            if current_thread_name:
+                current_thread_name = current_thread_name.split('::')[0]  # We want to see only the context
+                self.threadname = "%s::%X" % (current_thread_name, id(self))
+            else:
+                self.threadname = "anon-%X" % id(self)
 
         real_thread_no_greenlet = kwargs.pop('real_thread_no_greenlet', False)
         if is_module_patched("threading"):


idanarye:
This can easily explode:

```python
In [1]: import threading

In [2]: from easypy.concurrency import concurrent, MultiObject

In [3]: def recurse(depth):
   ...:     print('%s remain, I am %r' % (depth, threading.current_thread().name))
   ...:     if 0 < depth:
   ...:         with concurrent(recurse, depth - 1):
   ...:             pass
   ...:

In [4]: recurse(3)
3 remain, I am 'MainThread'
2 remain, I am 'MainThread::7F867069D5F8'
1 remain, I am 'MainThread::7F867069D5F8::7F867069DEF0'
0 remain, I am 'MainThread::7F867069D5F8::7F867069DEF0::7F8670842208'
```

We usually want only the first (context) and last (actual thread to filter on)

idanarye:
Also, this doesn't work for `MultiObject`:

```python
In [6]: MultiObject(range(5)).call(lambda n: print(n, threading.current_thread().name))
0 ThreadPoolExecutor-1_0
1 ThreadPoolExecutor-1_0
2 ThreadPoolExecutor-1_0
3 ThreadPoolExecutor-1_1
4 ThreadPoolExecutor-1_0
Out[6]: <MultiObject[NoneType] (x5/None)>
```

YuvalEvron:
> Also, this doesn't work for `MultiObject`:
> 
> ```python
> In [6]: MultiObject(range(5)).call(lambda n: print(n, threading.current_thread().name))
> 0 ThreadPoolExecutor-1_0
> 1 ThreadPoolExecutor-1_0
> 2 ThreadPoolExecutor-1_0
> 3 ThreadPoolExecutor-1_1
> 4 ThreadPoolExecutor-1_0
> Out[6]: <MultiObject[NoneType] (x5/None)>
> ```

This looks like a problem on `MultiObject`'s side. We need to find out why there is not thread name, but this should be a different PR.

koreno:
> > Also, this doesn't work for `MultiObject`:
> > ```python
> > In [6]: MultiObject(range(5)).call(lambda n: print(n, threading.current_thread().name))
> > 0 ThreadPoolExecutor-1_0
> > 1 ThreadPoolExecutor-1_0
> > 2 ThreadPoolExecutor-1_0
> > 3 ThreadPoolExecutor-1_1
> > 4 ThreadPoolExecutor-1_0
> > Out[6]: <MultiObject[NoneType] (x5/None)>
> > ```
> 
> This looks like a problem on `MultiObject`'s side. We need to find out why there is not thread name, but this should be a different PR.

MultiObject uses python's thread-pool framework, which controls the name. We could do that as long as we don't need to monkey-patch anything - I don't want to deal with more breakages due to python version differences.

easypy/concurrency.py
```diff
@@ -933,12 +933,19 @@ def __init__(self, func, *args, **kwargs):
...
-        self.threadname = kwargs.pop('threadname', 'anon-%X' % id(self))
         self.stopper = kwargs.pop('stopper', threading.Event())
         self.sleep = kwargs.pop('sleep', 1)
         self.loop = kwargs.pop('loop', False)
         self.timer = None
         self.console_logging = kwargs.pop('console_logging', True)
+        self.threadname = kwargs.pop('threadname', None)
+        if not self.threadname:
+            current_thread_name = threading.current_thread().name
+            if current_thread_name:
```
koreno: is there a thread that _doesn't_ have a name?
I think every `concurrent` call is going to have a `MainThread::` prefix...

YuvalEvron: Don't remember seeing something like this, but wanted to be on the safe side.

koreno: Well I think the `MainThread::` prefix is redundant and shouldn't be there. A `concurrent` thread spawned by the main thread should just have its `anon-xxx` name with no prefix.
would you fix that for us?

PR #199, opened 2019-10-20
timing/units: added a TimeInterval class

koreno:
and support for comparing Timers with time-based types

### DIFF
--- a/easypy/timing.py
+++ b/easypy/timing.py
@@ -6,6 +6,94 @@ from .decorations import parametrizeable_decorator
 from .units import Duration
 
 
+inf = float("inf")  # same as math.inf, but compatible with 3.4
+
+
+def to_timestamp(t):
+    return t.timestamp() if isinstance(t, datetime) else t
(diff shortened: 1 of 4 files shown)


easypy/timing.py
```diff
@@ -1,11 +1,97 @@
...
+    __slots__ = ("t0", "t1")
+
+    NO_START = -inf
+    NO_END = inf
+
+    def __init__(self, from_time=NO_START, to_time=NO_END):
+        self.t0 = self.NO_START if from_time is None else to_timestamp(from_time)
+        self.t1 = self.NO_END if to_time is None else to_timestamp(to_time)
+
+    @property
```
YuvalEvron: Why not `@safe_property`?

koreno: we don't use it anywhere else in `easypy`, so I don't see a reason to start specifically in this commit.

I think property functions should be trivial enough not to require this added complication and dependency.

PR #207, opened 2019-11-06
Add `done()` function to `concurrent` to check if the thread is finished

idanarye:
I used `done()` and made it a function instead of property because that's how it looks in `Futures`, but I'm not sure I like that name and if there are other preferences I can change it.

### DIFF
--- a/easypy/concurrency.py
+++ b/easypy/concurrency.py
@@ -1021,6 +1021,12 @@ class concurrent(object):
             raise self.exc
         return self._result
 
+    def done(self):
+        """
+        Return ``True`` if the thread is done (successfully or not)
+        """
+        return hasattr(self, '_result') or getattr(self, 'exc', None) is not None
+
     @contextmanager
     def paused(self):
         self.stop()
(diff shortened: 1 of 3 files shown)


easypy/concurrency.py
```diff
@@ -1021,6 +1021,12 @@ def result(self, timeout=None):
             raise self.exc
         return self._result
 
+    def done(self):
```
YuvalEvron: I think `is_done` is more appropriate name. Also, it can be a property

idanarye: If it's `is_done` it should remain a function. If it's a property it should be just `done`.

idanarye: Because "is" is kind of a verb...

YuvalEvron: I disagree, I have seen many properties which start with "is_", makes it clearer that this is a boolean value. I'll leave it for you to decide.

koreno:
Yes, let's follow the `Futures` pattern, as with `result()`.

Issue #212, opened 2020-01-09
concurrent swallows KeyboardInterrupt, which is not cool in gevent


koreno:
Our `concurrent` object seems to have had this bug for long-long time:

```python
    def _logged_func(self):
        try:
            ...
        except (KeyboardInterrupt, Exception) as exc:
            _logger.silent_exception("Exception in thread running %s: %s (traceback can be found in debug-level logs)", self.func, type(exc))
            self.exc = exc
            try:
                exc.timestamp = time.time()
            except Exception:
                pass
```

I think the original idea was to catch thread exceptions in order to log them (since the default behavior is to print them to stdout). But in gevent world this means the exception never travels to the main greenlet, effectively suppressing a SIGINT or Ctrl-C.

(For other `Exception`s gevent's behavior is to just print it - http://www.gevent.org/api/gevent.hub.html#gevent.hub.Hub.handle_error)

PR #213, opened 2020-01-09
concurrency+exceptions: don't swallow KeyboardInterrupt under gevent (#212); Also...

koreno:
apply_timestamp to exception only if missing


idanarye:
Why can't we just **always** propagate it? Because with regular threads the main thread also gets the exception?

koreno:
Yes, real threads don't actually receive the signal anyway.
We don't want to propagate it to prevent "spamming" stderr

PR #214, opened 2020-01-19
Improve `RecursionError` for aliasing-induced infinite recursion

### LINKED ISSUE #210: Improve infinite recursion exception when aliasing
Aliasing should catch the `RecursionError` and emit a better error that tells you about the object, the aliased object, and the attribute.

idanarye:
Fixes #210

### DIFF
--- a/tests/test_aliasing.py
+++ b/tests/test_aliasing.py
@@ -41,3 +41,24 @@ def test_aliasing_inherit():
     f = Foo("5")
     assert f.get("a") == 1
     assert f == 5
+
+
+def test_aliasing_infinite_recursion_exception():
+    @aliases('bar', static=False)
+    class Foo:
+        def __init__(self):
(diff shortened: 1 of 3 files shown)


tests/test_aliasing.py
```diff
@@ -41,3 +41,24 @@ def __init__(self, x):
...
+    class Bar:
+        def __init__(self, foo):
+            self.foo = foo
+
+        def baz(self):
+            return self.foo.baz()
+
+    with pytest.raises(RecursionError) as e:
+        Foo().baz()
+    assert str(e.value) == "Infinite recursion trying to alias 'baz' on Foo() (via Foo.bar)"
```
YuvalEvron: Shouldn't tis line be inside the context manager?

idanarye: No, because the exception is not thrown yet inside the context manager.

PR #216, opened 2020-01-21
contexts: added 'breakable_section'

koreno:
What do you think about this?
Kinda like "goto", but more pythonic...

In retrospect there were many situations where I used a closure/function just for its `return` capability ("quick" exit out of deep nesting). I think I'd prefer this construct instead.


idanarye:
Is `easypy.contexts` the right module though? It should have functions/classes that **deal** with contexts, not that **are** contexts.

Maybe this should go in a new module - `easypy.flowcontrol`?

YuvalEvron:
Honestly, I can't imagine myself using this feature.

koreno:
> Is `easypy.contexts` the right module though? It should have functions/classes that **deal** with contexts, not that **are** contexts.
> 
> Maybe this should go in a new module - `easypy.flowcontrol`?

I'm don't want to create a new module just for this... `contexts` also has KeyedStack which is a context-manager itself.
Besides, this is a "context which can be safely broken out of".

PR #217, opened 2020-01-22
Explicit signal declaration with function decoration

idanarye:
This is the beginning of an effort to make signals explicit. If we can integrate explicit signals and deprecate the implicit imports from `easypy.signals` that would be great.

For now this mechanism forbids multiple declaration of the same signal name if the signal was declared with a decorator. So, two signals with the same name cannot be declared with a decorator, and you can't mix the styles, but you can import the same signal twice (like you can today). Once we remove the implicit signal registration this limitation can be lifted and we can let Python's regular namespacing rules manage the signal names.


koreno:
I never saw this as a problematic limitation.
the idea of signals is that consumers and publishers don't need to know exactly where signals are declared (in which module). I.e, the namespace of a signal should not be a module.

It is true that as it is now, signals have a single global namespace - maybe that's something we should change.
Refer to https://github.com/getslash/gossip - where the namespaces are implemented as "hook groups"

koreno:
as for "implicit imports" - are you referring to the create-by-import pattern?
This "magic" is now standardized in python 3.7 - https://www.python.org/dev/peps/pep-0562/ - so I wouldn't view it negatively.

koreno:
at first I thought it was a bit awkward to have a noop function, but I do like it having a signature and validating it... nice!

idanarye:
My concern is not with namespacing - I only mentioned namespacing because it'll allow us to lift the restriction on duplicate signal names. My concern is with typos that become silent bugs.

Today we had a problem when someone created a method named `on_test_finished` instead of `on_test_finish`, and caused the runs to sometimes fail on `Can only create Signal objects from the MainThread` when the object containing that method is created in a thread. The problem is not the registration in thread - that only _exposed_ the real problem. The problem is that when the object is first registered from the main thread, it doesn't fail - it simply creates a new signal which **does not** get called when the test finished.

Explicit is better than implicit, and build errors (or their equivalent in interpreted languages - errors during source file import) - are better than logical bugs.

koreno:
I see.
Maybe we should start by disabling the ability to generate a new signal
from a class method altogether. That was never intended, I've never
considered it before.
As for the PR - so you're saying this new signal creation method is
incompatible with the previous? do you need them to live side-by-side? what
do we need to solve exactly?


- Ofer

idanarye:
Disabling the the ability to generate new signals from class methods would mean that if you have this:

```python
class Foo:
    def on_bar(self): ...

    
def baz():
    from easypy.signals import on_bar
    on_bar()
```

Then this won't work:
```python    
foo = Foo()
register_object(foo)
baz()
```
But this will
```python    
foo = Foo()
baz()
register_object(foo)
baz()
```

Even though the first `baz()` does nothing. If the `on_bar` must be created before it can be used as a method, this this constraint should be represented by `on_bar` the method depending on `on_bar` the signal.

> As for the PR - so you're saying this new signal creation method is 
> incompatible with the previous? do you need them to live side-by-side? what
> do we need to solve exactly?

They are only incompatible because I made them incompatible - I [created exceptions](https://github.com/weka-io/easypy/pull/217/commits/c3100aa62b5381f3dfb1269cd4f93492201797a8#diff-998c5dbcd254345295d87727999647b7R29-R38) which get raised if the two creation methods are used for the same signal name. I wanted to avoid making the order of declaration significant - if the signal gets imported first it doesn't get a signature but if it gets created with a decorator first it gets one. I could plant the signature in the existing signal, but then what will happen to the handlers that were registered before that? It can all be solved, but it doesn't seem worth the trouble seeing how this is a temporary restriction until we deprecate the import method.

My original intent was to first push this, and then work on making the handler registration explicit too - something like:

```python
@signal
def on_foo(): ...


class Bar:
    @on_foo.mark_handler  # may need a better name for this function
    def do_something(self):
	...
```
And then `register_object(Bar())` will recognize that it needs to register `do_something` as an `on_foo` signal.

But until we do that, I thought, methods will still be registered as handler by name, even for signals created by decoration, so they need to be written into `Signal.ALL` which means they can conflict with signals created by import. So I had to put that restriction.

But now that we discuss this, I think maybe I should just implement `mark_handler` in this PR, not register newstyle signals in `Signal.ALL`, and lift that restriction. Then we could convert signals one by one, both declaration and handler registration for each signal.

koreno:
got it, thanks for the clarifications

> ```python
>     @on_foo.mark_handler  # may need a better name for this function
> ```

`@on_foo.handler` works for me.

It would be nice to provide a `SignalHandlersMixin` class that would support the current behavior, where these decorators are not required. perhaps also handle registering objects when they are done instantiated+initialized

easypy/signals.py
```diff
@@ -146,11 +158,15 @@ class Signal:
 
     ALL = {}
 
-    def __new__(cls, name, asynchronous=None, swallow_exceptions=False, log=True):
+    def __new__(cls, name, asynchronous=None, swallow_exceptions=False, log=True, signature=None):
         try:
-            return cls.ALL[name]
+            signal = cls.ALL[name]
```
YuvalEvron: I think we don't need try/except here. `if name in cls.ALL` is more elegant.

easypy/signals.py
```diff
@@ -215,7 +235,41 @@ def registered(self, func, **kwargs):
         finally:
             self.unregister(func)
 
+    def _verify_handler_func(self, handler_func):
+        handler_signature = inspect.signature(handler_func)
+        bad_parameters = ', '.join(
+            '%s is %s' % (p, p.kind)
+            for p in handler_signature.parameters.values()
+            if p.kind in (p.POSITIONAL_ONLY, p.VAR_POSITIONAL))
```
YuvalEvron: Why don't we like positional params?

easypy/signals.py
```diff
@@ -335,6 +389,47 @@ def run_handler(handler):
         self.remove_handlers_if_exist(handlers_to_remove)
 
 
+@parametrizeable_decorator
+def signal(func, asynchronous=None, swallow_exceptions=False, log=True):
```
YuvalEvron: Why `asynchronous=None` and not `asynchronous=False`?

easypy/signals.py
idanarye: We don't like position **only** parameters. This is [a new feature in Python 3.8](https://docs.python.org/3/whatsnew/3.8.html#positional-only-parameters) (though builtin functions always had it):
```python
def foo(a, b, /, c, d): ...
```
`a` and `b` - the parameters before the `/` - are positional only, so you can't call them by name. You can do `foo(1, 2, 3, 4)` or `foo(1, 2, c=3, d=4)` but not `foo(a=1, b=2, c=3, d=4)`.

Since signal handlers get all their arguments by name, we'll need to do extra work to support positional arguments - and the RoI for that is too low, so it's better to just disallow them.

easypy/signals.py
idanarye: `asynchronous=False` means you are not **allowed** to put asynchronous handlers on the signal. `asynchronous=None` means both sync and async handlers are OK.

YuvalEvron: > `asynchronous=False` means you are not **allowed** to put asynchronous handlers on the signal. `asynchronous=None` means both sync and async handlers are OK.

It is very unclear. Is it documented somewhere?

easypy/signals.py
idanarye: This is the superior style. `try:` clause to get the signal, `except KeyError:` clause the handle the case it does not exist, `else:` clause to handle the case it does exist.

easypy/signals.py
idanarye: I'm just replicating the behavior of the `Signal` class...

idanarye:
I've added a another commit, that separates the namespaces and adds the `Signal.handler` decorator.

idanarye:
> It would be nice to provide a `SignalHandlersMixin` class that would support the current behavior, where these decorators are not required. perhaps also handle registering objects when they are done instantiated+initialized

If decorators are not required, then importing the signal is not required, and once again we reach the problem where typos in signals emit no errors.

PR #218, opened 2020-01-24
caching: added support for cache-hints via timecache.return_value

koreno:


koreno:
I'm not quite loving the `return_value`, maybe it should just be a `return timecache.WithHints(value, skip_cache=True)`?

PR #219, opened 2020-01-27
contexts: added 'breakable_section'

koreno:


idanarye:
How is this different from #216?

koreno:
It's got tests, and the branch name is fixed

PR #221, opened 2020-02-11
Add `ensure_same_defaults`

idanarye:
As I suggested in #185 - here is my implementation of `ensure_same_defaults`.

### DIFF
--- a/easypy/decorations.py
+++ b/easypy/decorations.py
@@ -5,6 +5,9 @@ This module is about making it easier to create decorators
 from functools import wraps, partial, update_wrapper
 from operator import attrgetter
 from abc import ABCMeta, abstractmethod
+import inspect
+
+from easypy.exceptions import TException
 
 
 def parametrizeable_decorator(deco):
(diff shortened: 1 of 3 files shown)


easypy/decorations.py
```diff
@@ -138,3 +141,50 @@ def foo(self):
...
+    params_with_defaults = [
+        param for param in sot_signature.parameters.values()
+        if param.default is not param.empty
+        and param.name not in ignore]
+
+    def gen_mismatches(func):
+        signature = inspect.signature(func)
+        for sot_param in params_with_defaults:
+            param = signature.parameters.get(sot_param.name)
+            if param is None:
```
YuvalEvron: The first two `if` clauses can be merged into one.

YuvalEvron:
I still think this we don't need this feature.

easypy/decorations.py
koreno: I think it's clearer this way

PR #222, opened 2020-02-21
resilience: support returning exceptions captured by decorator

koreno:


easypy/resilience.py
```diff
@@ -264,11 +271,13 @@ def resilience(msg="ignoring error {type}", acceptable=Exception, unacceptable=(
 resilient.info = partial(resilient, log_level=logging.INFO)
 resilient.warning = partial(resilient, log_level=logging.WARNING)
 resilient.error = partial(resilient, log_level=logging.ERROR)
+resilient.CAPTURE = object()
```
idanarye: Aren't we using `easypy.tokens` for that those things?

easypy/resilience.py
```diff
@@ -241,19 +243,24 @@ def resilience(msg="ignoring error {type}", acceptable=Exception, unacceptable=(
...
+    from .bunch import Bunch
+
     if unacceptable is None:
         unacceptable = ()
     elif isinstance(unacceptable, tuple):
         unacceptable += UNACCEPTABLE_EXCEPTIONS
     else:
         unacceptable = (unacceptable,) + UNACCEPTABLE_EXCEPTIONS
+
+    container = Bunch(exception=None)
```
idanarye: I think this justifies its own class. That - or we can introduce a `Cell` class to easypy.

easypy/resilience.py
```diff
@@ -217,9 +218,10 @@ def resilient(func=None, default=None, **kw):
 
     @wraps(func)
     def inner(*args, **kwargs):
-        with resilience(**kw):
+        with resilience(**kw) as container:
             return func(*args, **kwargs)
-        return default  # we reach here only if an exception was caught and handled by resilience
+        # we reach here only if an exception was caught and handled by resilience
+        return container.exception if (default is resilient.CAPTURE) else default
```
idanarye: I'm not sure what your usecase is, but this looks like an API disaster. If someone is going to check if the return type is an exception, they should just use `try`...`except`. And if they don't, and just let the value get passed along - that's just going to create weird hard-to-debug bugs.

How about introducing a [result type](https://en.wikipedia.org/wiki/Result_type) instead?

koreno: well, I think in the end my use-case is a bit specific, and maybe deserve to be treated without adding this feature, since, as it seems, not obvious.
the use case is sending off a bunch of methods, concurrently, to acquire some piece of information. Some methods may fail, some may take too long. I just want to continue on as soon as one of them gets it, doesn't matter which.
I think I should just implement that into concurrency

idanarye: This does look like a job for result types then, but it's never a good idea to mix several error handling models.

Another option is to make this feature more generic - maybe add an `on_exception` parameter that accepts a lambda that gets the exception and decides the default value based on it:

```python
if on_exception:
    return on_exception(container.exception)
```

And then:

```python
@resilient(..., on_exception=lambda e: e)
def some_function(...): ...
```

This would easily fulfill your usecase and still be general enough to ensure the bug potential I described only gets triggered if someone intentionally triggers it.

